PR #19, opened 2024-09-18
feat: Make `OffTargetDetector.filter()` generic

### LINKED ISSUE #18: `OverlapDetector.filter()` should be made generic
`OverlapDetector.filter` should be made generic so we can filter arbitrary subclasses of `Primer` without mypy raising a flag

msto:
Closes #18

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -90,13 +90,13 @@ files = [
 
 [[package]]
 name = "build"
-version = "1.2.1"
+version = "1.2.2"
 description = "A simple, correct Python build frontend"
 optional = false
 python-versions = ">=3.8"
 files = [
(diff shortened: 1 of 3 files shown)


tfenne (approved):
One question ... does Python have the ability to declare a type variable as part of a function declaration?  In scala we'd inline that type variable, e.g.

```
def filter[P <: Primer](primers: List[P]) -> List[P]
```

and having the type variable right there makes it easier to see what's going on.  
```

clintval:
@tfenne not yet but PEP 718 is tracking that language feature:

https://peps.python.org/pep-0718/

If we do get this syntax, it could arrive as soon as Py3.13 (but likely Py3.14).

https://devguide.python.org/versions/

PR #23, opened 2024-09-18
fix: Pad target region by `max_amplicon_length`

### LINKED ISSUE #16: The design region should be padded by the complete max amplicon length, not half this length
Per discussion with Tim and Erin today

When padding the target region to create the design region, the design region should be padded by the complete max amplicon length. It is currently padded by half this length, which artificially constrains the search space. e.g., for a max amplicon length of 400 bp, it would be valid to have one primer situated 300 bp to the left of the target and one primer situated 50 bp to the right of the target. However, if we only consider the 200 bp flanking either side of the target, this pair would not be considered as a valid design.

### LINKED ISSUE #17: Validate that sufficient space exists around the target to design primers, before executing primer3
`primer3` will error out if there is not sufficient space to design a primer around the target (e.g. `assertion (len >= 3) failed`)

`prymer` should validate that the design region includes at least `max_amplicon_length` bases outside the target, and raise an error if not. 

(And possibly raise a warning if the design region includes only `2 * max_amplicon_length` bases, or some other small window)

msto:
Closes #16 
Closes #17 

This PR fixes the padding of the target region to be consistent with the original `fgprimer` implementation. (My bad 🙃 )

https://github.com/fulcrumgenomics/fgprimer/blob/6cf2542e927ced37dd0dce4c335de8dff07789c7/src/main/scala/com/fulcrumgenomics/primerdesign/primer3/Primer3.scala#L89-L93

Other changes
- Validate that the max amplicon length permits sufficient space around the target region for primers to be designed
- Renamed `region` to `design_region`, to better reflect the target/design region distinction
- Renamed `_pad_target_region` to `_create_design_region`, for the same reason, and clarified its documentation
- Added `min_primer_length` property to `Primer3Parameters` (`max_primer_length` already existed)

### DIFF
--- a/prymer/primer3/primer3.py
+++ b/prymer/primer3/primer3.py
@@ -372,12 +372,13 @@ class Primer3(ExecutableRunner):
                 f"terminated, return code {self._subprocess.returncode}"
             )
 
-        region: Span = self._pad_target_region(
+        design_region: Span = self._create_design_region(
             target=design_input.target,
             max_amplicon_length=design_input.params.max_amplicon_length,
+            min_primer_length=design_input.params.min_primer_length,
         )
 
-        soft_masked, hard_masked = self.get_design_sequences(region)
+        soft_masked, hard_masked = self.get_design_sequences(design_region)
         global_primer3_params = {
             Primer3InputTag.PRIMER_FIRST_BASE_INDEX: 1,
             Primer3InputTag.PRIMER_EXPLAIN_FLAG: 1,
@@ -386,7 +387,7 @@ class Primer3(ExecutableRunner):
 
         assembled_primer3_tags = {
             **global_primer3_params,
-            **design_input.to_input_tags(design_region=region),
+            **design_input.to_input_tags(design_region=design_region),
         }
 
         # Submit inputs to primer3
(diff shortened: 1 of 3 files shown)


prymer/primer3/primer3.py
```diff
@@ -79,8 +79,8 @@
 ```python
 >>> for failure in left_result.failures: \
     print(failure)
-Primer3Failure(reason=<Primer3FailureReason.HIGH_TM: 'high tm'>, count=171)
-Primer3Failure(reason=<Primer3FailureReason.GC_CONTENT: 'GC content failed'>, count=26)
+Primer3Failure(reason=<Primer3FailureReason.HIGH_TM: 'high tm'>, count=406)
+Primer3Failure(reason=<Primer3FailureReason.GC_CONTENT: 'GC content failed'>, count=91)
```
msto, line 83: These needed to be updated after the change to padding increased the search space (and correspondingly increased the number of designed primers)

prymer/primer3/primer3.py
```diff
@@ -710,21 +711,38 @@ def _build_failures(
...
-        When the max amplicon length is odd, the left side of the target region will be padded with
-        one more base than the right side.
         """
+        padding: int = max_amplicon_length - target.length
         contig_length: int = self._dict[target.refname].length
-        padding_right: int = max(0, int((max_amplicon_length - target.length) / 2))
-        padding_left: int = max(0, max_amplicon_length - target.length - padding_right)
 
-        region: Span = replace(
+        if padding < min_primer_length:
```
tfenne, line 736: I think you want to move this check below creation of the `design_region` and check to ensure that the design_region start/end are at least min_primer_length </> the target's.  Otherwise it won't catch the case where the padding gets limited by the contig start/end.

tfenne (approved):
Approved once the ValueError check gets moved further south.

tfenne (approved):
LGTM - thanks @msto

PR #25, opened 2024-09-18
fix: Executable path validation tests should not be dependent on the user environment

### LINKED ISSUE #20: `test_validate_executable_path` fails locally
```py
=================================================================================== FAILURES ====================================================================================
_________________________________________________________________________ test_validate_executable_path _________________________________________________________________________

    def test_validate_executable_path() -> None:
        exec = "yes"
        exec_full_str = f"/usr/bin/{exec}"
        exec_full_path = Path(exec_full_str)
        assert exec_full_path.is_absolute()

        # find it on the PATH
>       assert exec_full_path == ExecutableRunner.validate_executable_path(executable=exec)
E       AssertionError: assert PosixPath('/usr/bin/yes') == PosixPath('/opt/homebrew/opt/coreutils/libexec/gnubin/yes')
E        +  where PosixPath('/opt/homebrew/opt/coreutils/libexec/gnubin/yes') = <bound method ExecutableRunner.validate_executable_path of <class 'prymer.util.executable_runner.ExecutableRunner'>>(executable='yes')
E        +    where <bound method ExecutableRunner.validate_executable_path of <class 'prymer.util.executable_runner.ExecutableRunner'>> = ExecutableRunner.validate_executable_path

tests/util/test_executable_runner.py:39: AssertionError
```

msto:
Closes #20 

Homebrew installation of the GNU coreutils caused a conflict with expectation in the `ExecutableRunner` validation tests.

Tests were updated to clear the `PATH` before executing.

### DIFF
--- a/tests/util/test_executable_runner.py
+++ b/tests/util/test_executable_runner.py
@@ -41,8 +41,8 @@ def test_validate_executable_path_not_executable() -> None:
 def test_validate_executable_path(tmp_path: Path) -> None:
     """
     `validate_executable_path` should return the `yes` executable in the following scenarios:
-    1. When the name of the executable is passed as a string.
-    2. When the absolute path to the executable is passed, either as a string or a Path.
+    1. The name of the executable is passed as a string, and the executable is on the user's PATH.
+    2. The absolute path to the executable is passed, either as a string or a Path.
     """
     expected_path = tmp_path / "yes"
     expected_path.touch()  # create the file
--- a/tests/util/test_executable_runner.py
+++ b/tests/util/test_executable_runner.py
@@ -38,21 +38,24 @@ def test_validate_executable_path_not_executable() -> None:
             ExecutableRunner.validate_executable_path(executable=tmpfile.name)
 
 
-@pytest.mark.parametrize("executable", ["yes", "/usr/bin/yes", Path("/usr/bin/yes")])
-def test_validate_executable_path(executable: str | Path) -> None:
+def test_validate_executable_path(tmp_path: Path) -> None:
     """
     `validate_executable_path` should return the `yes` executable in the following scenarios:
     1. When the name of the executable is passed as a string.
     2. When the absolute path to the executable is passed, either as a string or a Path.
     """
-    expected_path: Path = Path("/usr/bin/yes")
+    expected_path = tmp_path / "yes"
+    expected_path.touch()  # create the file
+    expected_path.chmod(755)  # make it executable
 
     # Clear the PATH, in case the user has a local version of `yes` elsewhere on their PATH
     with mock.patch.dict(os.environ, clear=True):
-        os.environ["PATH"] = "/usr/bin"
+        os.environ["PATH"] = str(tmp_path)
 
-        validated_path: Path = ExecutableRunner.validate_executable_path(executable=executable)
-        assert validated_path == expected_path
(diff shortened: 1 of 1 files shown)


tests/util/test_executable_runner.py
```diff
@@ -29,22 +30,32 @@ def test_validate_executable_path_not_executable() -> None:
...
+    """
+    expected_path: Path = Path("/usr/bin/yes")
 
-    # find it on the PATH
-    assert exec_full_path == ExecutableRunner.validate_executable_path(executable=exec)
-    # find it given an absolute path as a string
-    assert exec_full_path == ExecutableRunner.validate_executable_path(executable=exec_full_str)
-    # find it given an absolute path as a Path
-    assert exec_full_path == ExecutableRunner.validate_executable_path(executable=exec_full_path)
+    with mock.patch.dict(os.environ):
```
msto, line 42: NB: I'm using `unittest.mock` instead of `pytest.mocker` because of a bug/shortcoming in `pytest`. 

`patch.dict` only works as a context manager with `os.environ` if you use the standard library, it's not supported in `pytest`'s implementation

tests/util/test_executable_runner.py
```diff
@@ -29,22 +30,43 @@ def test_validate_executable_path_not_executable() -> None:
...
+
+
+def test_validate_executable_finds_on_path() -> None:
+    """
+    `validate_executable_path` should find executables on the PATH when the name of the executable
+    is passed as a string.
+    """
+    expected_path: Path = Path("/usr/bin/yes")
+
+    with mock.patch("shutil.which", return_value=expected_path.as_posix()):
```
msto, line 56: NB: I briefly tried monkeying around with patching `os.environ["PATH"]`, but `shutil.which` was still returning results based on the value of the PATH at the time 

I think the mock here is sufficient

tfenne:
@geoffjentry Could you review this one instead of me please?

tests/util/test_executable_runner.py
geoffjentry: OOI do you see it as better/worse/indifferent to go the `clear=True` route and just set the `PATH` in `os.environ`? 

It feels like it'd be a cleaner test to do it that way but a) I can see arguments against it and b) perhaps there's a reason I didn't see why that's not really feasible

geoffjentry (approved):

tests/util/test_executable_runner.py
msto: Ok I just updated to that effect. Dunno what blocked me there yesterday 🙃

tests/util/test_executable_runner.py
```diff
@@ -18,47 +18,55 @@ def test_close_twice() -> None:
...
-        with pytest.raises(ValueError, match="is not executable"):
-            ExecutableRunner.validate_executable_path(executable=exec_str)
-        # make it executable and test again
-        os.chmod(exec_str, 755)
-        exec_full_path: Path = exec_path.absolute()
-        assert exec_full_path == ExecutableRunner.validate_executable_path(executable=exec_str)
-        assert exec_full_path == ExecutableRunner.validate_executable_path(executable=exec_path)
-        assert exec_full_path == ExecutableRunner.validate_executable_path(
-            executable=exec_full_path
-        )
```
msto, line 64: this was redundant with the above tests

geoffjentry (approved):

tfenne (approved):

Issue #24, opened 2024-09-18
The `publish` badge on the home page is marked as failing


msto:
Two issues here, I think

1. The badget links to a search for an action named "Python package", of which we have none.

https://github.com/fulcrumgenomics/prymer/actions?query=workflow%3A%22Python+package%22

2. I assume the badge is meant to be linked to the "publish" action, but this action also appears to be broken (at least, the most recent execution failed, and so I'm assuming Nils manually published)

https://github.com/fulcrumgenomics/prymer/actions/workflows/publish.yml

clintval:
Duplicate of https://github.com/fulcrumgenomics/prymer/issues/3

I think I know what's happening. I believe the trust relationship was copied verbatim from pybedlite when we drafted https://github.com/fulcrumgenomics/pybedlite/pull/37. I'll co-assign myself and see if I can't figure it out.

Issue #3, opened 2024-08-21
pypi publishing tries to upload pybedlite


nh13:
See publish action: https://github.com/fulcrumgenomics/prymer/actions/runs/10494834616/job/29072084018

PR #26, opened 2024-09-18
chore: fix PyPi trusted publishing in GitHub Actions

clintval:
Closes: https://github.com/fulcrumgenomics/prymer/issues/3
Closes: https://github.com/fulcrumgenomics/prymer/issues/24

And the updated PyPi trusted publisher config:

Screenshot 2024-09-18 at 11 25 30 AM

### DIFF
--- a/.github/workflows/publish_prymer.yml
+++ b/.github/workflows/publish_prymer.yml
@@ -1,4 +1,4 @@
-name: publish
+name: publish prymer
 
 on:
   push:
@@ -124,7 +124,6 @@ jobs:
         uses: softprops/action-gh-release@v2
         with:
           name: ${{ github.ref_name }}
-          body: |
-            ${{ needs.draft-changelog.outputs.release_body }}
+          body: ${{ needs.make-changelog.outputs.release_body }}
           draft: false
           prerelease: false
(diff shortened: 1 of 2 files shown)


clintval:
This will not be possible to test without making a release of some kind.

clintval:
Very confused why tests are failing...

.github/workflows/publish_prymer.yml
```diff
@@ -1,4 +1,4 @@
-name: publish
+name: publish prymer
```
msto, line 1:
```suggestion
name: Publish prymer
```

**nit** I like to use title case so it renders a bit more nicely in the checks UI

msto (approved):
lgtm! thanks for fixing

.github/workflows/publish_prymer.yml
clintval: > **nit** I like to use title case so it renders a bit more nicely in the checks UI

I agree! In this case I'm matching the repo's pattern of having them be lowercase so this one doesn't appear out of place.

Screenshot 2024-09-19 at 8 34 49 AM

tfenne (approved):

Issue #30, opened 2024-09-19
Update README to refer to bioconda package for `bwa-aln-interactive`


ameynert:
https://anaconda.org/bioconda/bwa-aln-interactive

Issue #27, opened 2024-09-18
GitHub Action tests are using deprecated Mambaforge


clintval:
<img width="914" alt="Screenshot 2024-09-18 at 12 25 23 PM" src="https://github.com/user-attachments/assets/0e88a5f6-f152-4209-90ac-79c192639ffc">

You can see the warnings here: https://github.com/fulcrumgenomics/prymer/actions/runs/10927951278

Issue #5, opened 2024-08-21
Add `prymer` to bioconda


nh13:
See: https://github.com/bioconda/bioconda-recipes/pull/50249

msto:
The linked pull request was merged.

https://github.com/bioconda/bioconda-recipes/pull/50249

The bioconda recipe does not include `bwa-interactive` - I am not sure if we want to tie that to this issue, or close this and open another.

clintval:
I discovered the recipe on bioconda is shipping a _**non-interactive**_ version of bwa. See here for more information and next steps:

https://github.com/fulcrumgenomics/prymer/pull/32#issuecomment-2362821429

Issue #15, opened 2024-09-18
Error messages are not always propagated through `primer3_error`


msto:
`primer3_error()` fails to propagate errors when the logged error includes an equals sign.

When this occurs, the line is stored in `primer3_results` rather than `error_lines`. 

Could we add stricter validation/parsing of the primer3 output, so underlying error messages are visible?
```
# contents of `primer3_results`
{'/Users/msto/.miniforge-x86/envs/gia_primer_design/bin/primer3_core:libprimer3.cc:5745, assertion (len >': ' 3) failed'}

# corresponding logged line
/Users/msto/.miniforge-x86/envs/gia_primer_design/bin/primer3_core:libprimer3.cc:5745, assertion (len >= 3) failed
```

https://github.com/fulcrumgenomics/prymer/blob/8d332df27d1a467dd66369cc2e16183f033ab8d2/prymer/primer3/primer3.py#L402-L412

https://github.com/fulcrumgenomics/prymer/blob/8d332df27d1a467dd66369cc2e16183f033ab8d2/prymer/primer3/primer3.py#L423-L435

PR #32, opened 2024-09-20
feat: use bwa-aln-interactive and upgrade developer's docs

clintval:
Closes https://github.com/fulcrumgenomics/prymer/issues/5
Closes https://github.com/fulcrumgenomics/prymer/issues/30
Closes https://github.com/fulcrumgenomics/prymer/issues/27

This PR also:

- Addresses improper installation of poetry.
- Adds some polish to the main README and developer documentation.


clintval:
Local and remote tests hang forever when you install `bwa-aln-interactive` from bioconda and run tests.

The built executable from bioconda is either broken or this library isn't compatible with it yet.

Screenshot 2024-09-19 at 7 33 29 PM

clintval:
I've narrowed the issue down to this "hack" which does not appear to work for me:

https://github.com/fulcrumgenomics/prymer/blob/e942368ca0e28d5f0533c6fc5a6b6b35b59cf764/prymer/offtarget/bwa.py#L287-L300

clintval:
I figured it out. The GitHub release of `bwa-aln-interactive` is broken and we are shipping the non-interactive bwa. I'll next fix up the public repository and the bioconda package by incrementing the `build` number.

As a side quest, I was able to remove the `AlignmentFile` "hack"!

msto:
> I wish prymer had better logging redirects for the underlying code it calls. Debugging this was painful without knowing where all the standard errors go (hint... it's /dev/null and the user can't change that).

Can you open a follow-up issue for this? I agree

clintval:
> > I wish prymer had better logging redirects for the underlying code it calls. Debugging this was painful without knowing where all the standard errors go (hint... it's /dev/null and the user can't change that).
> 
> Can you open a follow-up issue for this? I agree

You have one open for Primer3:

- https://github.com/fulcrumgenomics/prymer/issues/15

I wrote one up for bwa:

- https://github.com/fulcrumgenomics/prymer/issues/41

prymer/offtarget/bwa.py
```diff
@@ -297,10 +306,9 @@ def __init__(
 
     def __signal_bwa(self) -> None:
         """Signals BWA to process the queries"""
-        for _ in range(3):
-            self._subprocess.stdin.flush()
-            self._subprocess.stdin.write("\n\n")
-            self._subprocess.stdin.flush()
+        self._subprocess.stdin.flush()
+        self._subprocess.stdin.write("\n" * 16)
+        self._subprocess.stdin.flush()
```
clintval, line 311: The Bioconda built executable requires more newlines before emitting an alignment.

Why the executable (regardless of compilation environment) needs more than 2 newlines is currently a mystery.

.github/workflows/tests.yml
```diff
@@ -19,87 +19,46 @@ jobs:
...
 
       - name: Set up Python ${{ matrix.PYTHON_VERSION }}
         uses: actions/setup-python@v5
         with:
           python-version: ${{ matrix.PYTHON_VERSION }}
 
+      - name: Install poetry
+        run: |
+          python -m pip install --upgrade pip
+          python -m pip install poetry==${{env.POETRY_VERSION}}
```
msto, line 31: **suggestion** Should we use `pipx` to install `poetry` so it gets its own environment?

prymer/offtarget/bwa.py
```diff
@@ -297,10 +306,9 @@ def __init__(
 
     def __signal_bwa(self) -> None:
         """Signals BWA to process the queries"""
-        for _ in range(3):
-            self._subprocess.stdin.flush()
-            self._subprocess.stdin.write("\n\n")
-            self._subprocess.stdin.flush()
+        self._subprocess.stdin.flush()
+        self._subprocess.stdin.write("\n" * 16)
```
msto, line 310: **suggestion** Leave a comment/breadcrumb to document the reason for this and the choice of 16

msto (approved):

clintval:
@msto's tested the developer docs and tests passed.

I think this is ready for review!

clintval:
https://github.com/coderabbitai full review

prymer/offtarget/bwa.py
```diff
@@ -296,11 +308,12 @@ def __init__(
...
     def __signal_bwa(self) -> None:
-        """Signals BWA to process the queries"""
-        for _ in range(3):
-            self._subprocess.stdin.flush()
-            self._subprocess.stdin.write("\n\n")
-            self._subprocess.stdin.flush()
+        """Signals BWA to process the queries."""
+        self._subprocess.stdin.flush()
+        # NB: the executable compiled on different platforms require a different number of newlines
+        # NB: it is not currently understood why this is, but 16 spaces seems to work for all tested
```
tfenne, line 314:
```suggestion
        # NB: it is not currently understood why this is, but 16 newlines seems to work for all tested
```

tfenne (approved):

PR #37, opened 2024-09-23
refactor: export OffTargetDetector

msto:
This should be in the `offtarget` namespace if `OffTargetResult` is

### DIFF
--- a/tests/api/test_picking.py
+++ b/tests/api/test_picking.py
@@ -26,7 +26,7 @@ from prymer.api.primer_pair import PrimerPair
 from prymer.api.span import Span
 from prymer.api.span import Strand
 from prymer.ntthal import NtThermoAlign
-from prymer.offtarget.offtarget_detector import OffTargetDetector
+from prymer.offtarget import OffTargetDetector
 
 
 @pytest.fixture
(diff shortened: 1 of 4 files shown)


tfenne:
I tend to agree .. but can you also search/replace to fix the imports in the examples please?

msto:
Yep, done. (And did the same for #36 )

tests/api/test_picking.py
```diff
@@ -10,23 +10,23 @@
...
-from prymer.api.minoptmax import MinOptMax
-from prymer.api.picking import FilteringParams
 from prymer.api.picking import _dist_penalty
 from prymer.api.picking import _seq_penalty
-from prymer.api.picking import build_and_pick_primer_pairs
-from prymer.api.picking import build_primer_pairs
 from prymer.api.picking import check_primer_overlap
 from prymer.api.picking import is_acceptable_primer_pair
-from prymer.api.picking import pick_top_primer_pairs
 from prymer.api.picking import score as picking_score
```
msto, line 27: Note that the remaining functions are not included in the `api` namespace, and must be imported from the appropriate submodule. 

IMO these functions should either be added to the `api` namespace, or marked as private with a leading underscore, but I don't think that needs to fall within the scope of this PR

tfenne (approved):

PR #44, opened 2024-09-24
feat: Add support for `min_primer_pair_hits` to `OffTargetDetector` 

### LINKED ISSUE #43: `OffTargetDetector` should relax the constraint that each primer pair have at least one hit in the reference
Adapted from our Slack discussion:

## Context

I am running into an issue with `OffTargetDetector`, and I am not certain whether it is a bug or an oddity of my use case.

As I've mentioned, I am designing primer pairs to amplify the junctions of a donor integration into a host genome.

Upstream of the primer design, I have assembled the expected integration sequence, consisting of the linearized donor flanked by 500 bp of host genome on either side of the expected insertion site.

```
           [========= 500 bp host sequence =======][~~~~~~~~~~ linearized donor ~~~~~~~~~~][========= 500 bp host sequence =======]
```           

`primer3` uses a FASTA containing this assembled sequence to design primers, and `build_primer_pairs()`  uses it to construct the amplicon sequence.

However, when searching for off-target hits, I am searching the complete reference genome, which does not include the donor. This is fine when searching for off-target hits of individual primers, but I am encountering an issue when evaluating primer pairs.

Specifically, `OffTargetDetector` requires that there be at least one amplicon among the pairs' hits, i.e. that each primer have at least one hit in the reference. This (ideally :slightly_smiling_face: ) does not happen when one of the primers is designed against a sequence that is not present in the reference.

https://github.com/fulcrumgenomics/prymer/blob/70f684086feb5f204e1f8fd659e7b7443cd4e845/prymer/offtarget/offtarget_detector.py#L273

I think it would be appropriate to relax the constraint to `0 <= len(amplicons)`.  It makes sense to me that the off-target detection could be done against an arbitrary reference

msto:
Closes #43 

This PR introduces a new parameter to `OffTargetDetector`, `min_primer_pair_hits`. This permits the user to relax the current constraint that every primer pair have at least one hit in the reference. 

I have set the default value to 1, so default behavior remains unchanged by this PR, and documented the use case where it may be appropriate to change this value.

### DIFF
--- a/prymer/offtarget/offtarget_detector.py
+++ b/prymer/offtarget/offtarget_detector.py
@@ -158,10 +158,11 @@ class OffTargetDetector:
             max_primer_pair_hits: the maximum number of amplicons a primer pair can make and be
                 considered passing
             min_primer_pair_hits: The minimum number of amplicons a primer pair can make and be
-                considered passing. The default is 1, which is appropriate when the primer pair is
-                being evaluated for off-target hits against the same reference genome from which the
-                primers were generated. If the primer pair was generated from a different reference
-                sequence, it may be appropriate to set this value to 0.
+                considered passing. (In most cases, this is the number of amplicons a primer pair is
+                expected to generate.) The default is 1, which is appropriate when the primer pair
+                is being evaluated for off-target hits against the same reference genome from which
+                the primers were generated. If the primer pair was generated from a different
+                reference sequence, it may be appropriate to set this value to 0.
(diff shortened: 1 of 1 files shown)


tfenne (approved):
Only question - and I'd love others' input - is whether it should be `min_` or `expected_`?

msto:
Per slack discussion and approval from @emmcauley and @ameynert, we are choosing `min_`.

I added a brief note to the docstring to indicate that the min is typically the number of amplicons a primer pair is expected to generate.

PR #46, opened 2024-09-24
doc: Expand the documentation for `OffTargetDetector`

msto:
I added documentation to explain how the class can be used to filter primers vs primer pairs, and the behavior of each filtering method. 

I also added `Args` and `Return` blocks to the public methods.

### DIFF
--- a/prymer/offtarget/offtarget_detector.py
+++ b/prymer/offtarget/offtarget_detector.py
@@ -358,10 +358,27 @@ class OffTargetDetector:
         return result
 
     def mappings_of(self, primers: list[PrimerType]) -> dict[str, BwaResult]:
-        """Function to take a set of primers and map any that are not cached, and return mappings
-        for all of them.  Note: the genomics sequence of the returned mappings are on the opposite
-        strand of that of the strand of the primer.  I.e. we map the complementary bases (reversed)
-        to that of the primer."""
+        """
+        Map primers to the reference genome using `bwa aln`.
+
+        Alignment results may be optionally cached for efficiency. Set `cache_results` to `True`
+        when instantiating the [[OffTargetDetector]] to enable caching.
+
+        Any primers without cached results, or all primers when `cache_results` is `False`, will be
+        mapped to the reference genome using `bwa aln` and the specified alignment parameters.
+
+        **Note**: The reverse complement of each primer sequence is used for mapping. The `query`
+        sequence in each `BwaResult` will match the input primer sequence, as will the sequences
+        used as keys in the output dictionary. However, the coordinates reported in each `BwaHit`
+        associated with a result will correspond to complementary sequence.
+
+        Args:
+            primers: A list of primers to map.
+
+        Returns:
+            A dictionary mapping each primer's sequence to its alignment results.
+            See `BwaResult` for details regarding the attributes included in each result.
+        """
 
         primers_to_map: list[PrimerType]
         if not self._cache_results:
--- a/prymer/offtarget/offtarget_detector.py
+++ b/prymer/offtarget/offtarget_detector.py
@@ -107,8 +107,9 @@ class OffTargetResult:
             pair would generate an acceptable number of amplicons in the reference genome (i.e.
             `min_primer_pair_hits <= num_amplicons <= max_primer_pair_hits`). False otherwise.
         cached: True if this result is part of a cache, False otherwise.  This is useful for testing
-        spans: the set of mappings of the primer pair to the genome or an empty list if mappings
-            were not retained
+        spans: The set of mappings of the primer pair to the genome (i.e., the region spanned by the
+            inferred amplicon). This list will be empty if the generating [[OffTargetDetector]] was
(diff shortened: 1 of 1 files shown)


prymer/offtarget/offtarget_detector.py
```diff
@@ -150,6 +165,19 @@ def __init__(
...
+        This class accepts a large number of parameters to control the behavior of the off-target
+        detection. The parameters are used in the various aspects of off-target detection as
+        follows:
+
+        1. Alignment (off-target hit detection): `ref`, `executable`, `threads`,
+           `three_prime_region_length`, `max_mismatches_in_three_prime_region`, `max_mismatches`,
+           and `max_primer_hits`.
+        2. Filtering of individual primers: `max_primer_hits`.
+        3. Checking of primer pairs: `max_primer_hits`, `min_primer_pair_hits`,
+           `max_primer_pair_hits`, and `max_amplicon_size`.
```
msto, line 179: I thought it'd be helpful to have the parameters grouped by the function(s) that use them. I feel like there could be a better way to organize/present this, lmk if you have any suggestions

prymer/offtarget/offtarget_detector.py
```diff
@@ -150,6 +165,19 @@ def __init__(
         executable: str | Path = "bwa",
     ) -> None:
         """
+        Initialize an [[OffTargetDetector]].
```
ameynert, line 168: Can the primer pair related parameters be given defaults? I don't need them for checking single primers.

ameynert (commented):
Just the one suggestion about default params. The extra docstrings are v helpful!

I'm also using this API a little differently as I'm looking across multiple FASTA files, so I'm calling `.mappings_of` directly and aggregating hit counts

prymer/offtarget/offtarget_detector.py
```diff
@@ -293,10 +357,27 @@ def _build_off_target_result(
...
+        Alignment results may be optionally cached for efficiency. Set `cache_results` to `True`
+        when instantiating the [[OffTargetDetector]] to enable caching. 
+
+        Any primers without cached results, or all primers when `cache_results` is `False`, will be
+        mapped to the reference genome using `bwa aln` and the specified alignment parameters. 
+
+        **Note**: The reverse complement of each primer sequence is used for mapping. The `query`
+        sequence in each `BwaResult` will match the input primer sequence, as will the sequences
+        used as keys in the output dictionary. However, the coordinates reported in each `BwaHit`
+        associated with a result will correspond to complementary sequence.
```
msto, line 372: Is this sufficiently clear?

prymer/offtarget/offtarget_detector.py
msto: I have been wondering something similar - if it's possible to provide defaults for more of these parameters, or if the features should be decoupled

cc @tfenne @emmcauley

prymer/offtarget/offtarget_detector.py
```diff
@@ -122,15 +124,28 @@ class OffTargetResult:
...
+    `OffTargetDetector` uses a [custom, interactive
+    version](https://github.com/fulcrumgenomics/bwa-aln-interactive/) of `bwa aln` to perform
+    off-target detection. This approach is faster and more sensitive than traditional isPCR and in
+    addition can correctly detect primers that are repetitive and contain many thousands or millions
+    of mappings to the genome.
 
     Note that while this class invokes BWA with multiple threads, it is not itself thread-safe.
     Only one thread at a time should invoke methods on this class without external synchronization.
+
+    Off-target detection may be performed for individual primers or primer pairs.
```
tfenne, line 139:
```suggestion
    Off-target detection may be performed for individual primers and/or primer pairs.
```

prymer/offtarget/offtarget_detector.py
tfenne: I agree this is helpful.  I'm happy to come back later and expand this further too.

prymer/offtarget/offtarget_detector.py
```diff
@@ -200,25 +228,61 @@ def __init__(
...
 
     def filter(self, primers: list[PrimerType]) -> list[PrimerType]:
-        """Filters an iterable of Primers to return only those that have less than
-        `max_primer_hits` mappings to the genome."""
+        """
+        Remove primers that have more than `max_primer_hits` mappings to the genome.
+
+        This method maps each primer to the specified reference with `bwa aln` to search for
+        off-target hits. Note that when the reference includes the sequence from which the primers
+        were design, the on-target hit will be included in the hit count. `max_primer_hits` should
```
tfenne, line 236:
```suggestion
        were designed, the on-target hit will be included in the hit count. `max_primer_hits` should
```

prymer/offtarget/offtarget_detector.py
tfenne: It is, but I think this function should be private.

tfenne (approved):

PR #48, opened 2024-09-25
feat: add probe-specific logic to Primer3

emmcauley:


prymer/api/probe.py
```diff
@@ -0,0 +1,22 @@
+from dataclasses import dataclass
+
+from fgpyo.util.metric import Metric
+
+from prymer.api.primer import Primer
+
+
+@dataclass(frozen=True, init=True, kw_only=True, slots=True)
+class Probe(Primer, Metric["Probe"]):
```
msto, line 9: **issue** Per discussion on the working group call today, we are going to add these fields directly to `Primer` (and maybe rename the class to `Oligo` or similar)

prymer/primer3/primer3.py
```diff
@@ -371,12 +366,16 @@ def design_primers(self, design_input: Primer3Input) -> Primer3Result:  # noqa:
...
-        design_region: Span = self._create_design_region(
-            target_region=design_input.target,
-            max_amplicon_length=design_input.primer_and_amplicon_params.max_amplicon_length,
-            min_primer_length=design_input.primer_and_amplicon_params.min_primer_length,
-        )
+        design_region: Span
+        match design_input.task:
+            case PickHybProbeOnly():
+                design_region = design_input.target
+            case _:
```
msto, line 373: **suggestion** consider requiring that the provided task is one of the three design tasks before executing this block, and keep the `case _` branch to raise an error for any other task

prymer/primer3/primer3.py
```diff
@@ -760,3 +793,30 @@ def _create_design_region(
...
+        else:
+            raise ValueError(f"Primer3 did not return the count tag: {count_tag}")
+    count: int = int(maybe_count)
+
+    return count
+
+
+def _has_acceptable_dinuc_run(
+    design_input: Primer3Input, oligo_design: Union[Primer, Probe]
+) -> bool:
```
msto, line 815:
```suggestion
) -> bool:
    """
    True if the design's longest dinucleotide run is no more than the stipulated maximum.
    
    For primer designs, the maximum is recorded in the input's `PrimerAndAmpliconParameters.primer_max_dinuc_bases`.
    
    For probe designs, the maximum is recorded in the input's `ProbeParameters.probe_max_dinuc_bases`.
```

**suggestion** Document what makes a dinucleotide run acceptable

(args/return blocks aren't necessary in private functions, but it's helpful to at least have a summary docstring)

msto (changes requested):

prymer/api/probe.py
emmcauley: I should have documented this because I just bumped up against why I defaulted them both to `None` in the first place-- if you specify the default weights for either task, then Primer3 uses those weights as part of your scoring function, and it will complain if the corresponding params are not there, because it has no delta-off-optimal to score.

PR #49, opened 2024-09-25
feat: add `Probe` object to wrap probe design results from Primer3

emmcauley:
Originally we wanted to add a `Probe` object that inherits from `Primer` and added additional attributes of a probe. We have decided to change course to refactor `Primer` and `Probe` into the same dataclass, `Oligo`. Standby for refactor.

### DIFF
--- a/prymer/api/probe.py
+++ b/prymer/api/probe.py
@@ -1,0 +1,22 @@
+from dataclasses import dataclass
+
+from fgpyo.util.metric import Metric
+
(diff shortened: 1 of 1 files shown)


msto (commented):
per yesterday's discussion I think this can be closed, right? We are going to add these fields to `Primer` (and possibly rename it)?

emmcauley:
@msto agreed -- I will close this PR and open a new one with additional context

PR #60, opened 2024-10-03
feat: add thermodynamic-related weights and parameters for `PickHybProbeOnly` task

emmcauley:
This PR adds thermodynamic-related weights and parameters to `ProbeWeights()` and `ProbeParameters()` respectively. 

There are 3 additions to the `ProbeParameters` class:
* `probe_max_self_any_thermo`
* `probe_max_self_end_thermo`
* `probe_max_hairpin_thermo`

If these are not provided, they are set in the `__post_init()__` to the defaults specified in the Primer3 manual (10 degrees lower than the minimal melting temperature of the probe. The range of allowable melting temperatures is a required argument).

There are 3 corresponding additions to the `ProbeWeights` class: `probe_wt_self_any_th`, `probe_wt_self_end_th`, and `probe_wt_hairpin_th`.

No other thermodynamic-related changes of weights or parameters have been made to the `PrimerAndAmpliconWeights` or `PrimerAndAmpliconParameters` classes.


prymer/primer3/primer3_parameters.py
```diff
@@ -186,6 +188,14 @@ class ProbeParameters:
...
     Please see the Primer3 manual for additional details: https://primer3.org/manual.html#globalTags
 
+    Primer3 uses both thermodynamic and alignment-based approaches to quantify oligo
+    self-complementarity. In general, these settings are meant to limit problematic oligo
+    self-complementarity and avoid primer-dimers or other nonspecific binding of probes.
+
+    If they are not provided, `probe_max_self_any_thermo`, `probe_max_self_end_thermo`, and
+    `probe_max_hairpin_thermo` will be set to default values as specified in the Primer3 manual.
+    The default value is 10 degrees lower than the minimal melting temperature specified for
+    probe design.
```
ameynert, line 198: Can we have both deltaG and Tm? I've been using deltaG as a filter.

The docstring here should be very clear that if a user sets any of these attributes, they have to be *absolute* values, even though the default is relative to the min probe Tm.

ameynert (commented):
I haven't used penalties at all, so can't comment on that.

prymer/primer3/primer3_parameters.py
```diff
@@ -204,6 +217,10 @@ def __post_init__(self) -> None:
             raise TypeError("Probe melting temperatures and GC content must be floats")
         if self.probe_max_dinuc_bases % 2 == 1:
             raise ValueError("Max threshold for dinucleotide bases must be an even number of bases")
+        default_thermo_max: float = self.probe_tms.min - 10.0
+        for field in fields(self):  # if *_thermo fields are None, set to default
+            if field.type is Optional[float] and getattr(self, field.name) is None:
+                object.__setattr__(self, field.name, default_thermo_max)
```
msto, line 223: **suggestion** I think we should explicitly check that a field is one of the `probe_max_*_thermo` fields before setting the max.

Not an issue right now but it's defensive against the introduction of other `Optional[float]` attributes

```suggestion
        thermo_max_fields = ["probe_max_self_any_thermo", "probe_max_self_any_thermo", "probe_max_hairpin_thermo"]
        for field in fields(self):
            if field.name in thermo_max_fields and getattr(self, field.name) is None:
                object.__setattr__(self, field.name, default_thermo_max)
```

(you could also do something like `field.name.startswith("probe_max_")` if that's sufficiently specific

tests/primer3/test_primer3_input.py
```diff
@@ -101,8 +101,7 @@ def test_probe_design_only_valid(
     )
     mapped_dict = test_input.to_input_tags(design_region=test_design_region)
     assert mapped_dict[Primer3InputTag.PRIMER_PICK_INTERNAL_OLIGO] == 1
-
-    assert len(mapped_dict.keys()) == 21
+    assert len(mapped_dict.keys()) == 27
```
msto, line 104: **note** you can apply `len` directly to a dict!

```suggestion
    assert len(mapped_dict) == 27
```

tests/primer3/test_primer3_weights.py
```diff
@@ -32,7 +32,10 @@ def test_probe_weights_valid() -> None:
     assert test_dict[Primer3InputTag.PRIMER_INTERNAL_WT_TM_GT] == 1.0
     assert test_dict[Primer3InputTag.PRIMER_INTERNAL_WT_GC_PERCENT_LT] == 0.5
     assert test_dict[Primer3InputTag.PRIMER_INTERNAL_WT_GC_PERCENT_GT] == 0.5
-    assert len((test_dict.values())) == 6
+    assert test_dict[Primer3InputTag.PRIMER_INTERNAL_WT_SELF_ANY_TH] == 1.0
+    assert test_dict[Primer3InputTag.PRIMER_INTERNAL_WT_SELF_END_TH] == 1.0
+    assert test_dict[Primer3InputTag.PRIMER_INTERNAL_WT_HAIRPIN_TH] == 1.0
+    assert len((test_dict.values())) == 9
```
msto, line 38: ditto!

```suggestion
    assert len(test_dict) == 9
```

prymer/primer3/primer3_parameters.py
emmcauley: thanks Alison! I've updated the docstring in the most recent commits 👍

prymer/api/oligo.py
```diff
@@ -95,6 +112,11 @@ class Oligo(OligoLike, Metric["Oligo"]):
         bases: the base sequence of the oligo (excluding any tail)
         tail: an optional tail sequence to put on the 5' end of the primer
         name: an optional name to use for the primer
+        self_any_th: an optional calculated melting temperature that represents
+            the tendency of an oligo to bind to itself (self-complementarity)
+        self_end_th: an optional calculated melting temperature that represents
+            the tendency of a primer to bind to the 3'-END of an identical primer
+        hairpin_th: an optional calculated melting temperature of the oligo hairpin structure
```
tfenne, line 119: What do others think about giving these names that are more descriptive vs. sticking with what primer3 calls them?  I would propose:

- `self_any_th` -> `homodimer_tm`
- `self_end_th` -> `3p_anchored_homodimer_tm`
- `hairpin_th` -> `secondary_structure_tm`

prymer/primer3/primer3.py
```diff
@@ -530,13 +530,27 @@ def _build_oligos(
...
                 Oligo(
                     bases=bases,
                     tm=float(design_results[f"PRIMER_{design_task.task_type}_{idx}_TM"]),
                     penalty=float(design_results[f"PRIMER_{design_task.task_type}_{idx}_PENALTY"]),
                     span=span,
+                    self_any_th=float(
+                        design_results.get(
+                            f"PRIMER_{design_task.task_type}_{idx}_SELF_ANY_TH", None
+                        )
+                    ),
```
tfenne, line 545: In my hands `float(None)` raises an exception.  I think this needs to account for that possibility.

tfenne: Relatedly, should we enforce `PRIMER_THERMODYNAMIC_OLIGO_ALIGNMENT=1` and any other necessary parameters when submitting in `primer3.py::design()` to ensure thermodynamic values are always produced?  It's the default for the latest version, but not all verisons.

prymer/primer3/primer3_parameters.py
```diff
@@ -186,6 +188,17 @@ class ProbeParameters:
...
     Please see the Primer3 manual for additional details: https://primer3.org/manual.html#globalTags
 
+    Primer3 uses both thermodynamic and alignment-based approaches to quantify oligo
+    self-complementarity. In general, these settings are meant to limit problematic oligo
+    self-complementarity and avoid primer-dimers or other nonspecific binding of probes.
+
+    If they are not provided, `probe_max_self_any_thermo`, `probe_max_self_end_thermo`, and
+    `probe_max_hairpin_thermo` will be set to default values as specified in the Primer3 manual.
+    The default value is 10 degrees lower than the minimal melting temperature specified for
+    probe design (i.e., when not provided, values are specified relative to the probe design).
```
tfenne, line 198:
```suggestion
    probe design (i.e. when not provided, values are specified relative to the probe design).
```

prymer/primer3/primer3_parameters.py
```diff
@@ -186,6 +188,17 @@ class ProbeParameters:
 
     Please see the Primer3 manual for additional details: https://primer3.org/manual.html#globalTags
 
+    Primer3 uses both thermodynamic and alignment-based approaches to quantify oligo
+    self-complementarity. In general, these settings are meant to limit problematic oligo
+    self-complementarity and avoid primer-dimers or other nonspecific binding of probes.
```
tfenne, line 193: I think I would re-word this to say something like "While Primer3 supports alignment based and thermodynamic methods for simulating hybridizations, prymer enforces the use of the thermodynamic approach which is slightly more computationally expensive but superior."

prymer/primer3/primer3_parameters.py
```diff
@@ -196,6 +209,9 @@ class ProbeParameters:
     probe_max_dinuc_bases: int = 4
     probe_max_polyX: int = 5
     probe_max_Ns: int = 0
+    probe_max_self_any_thermo: Optional[float] = None
+    probe_max_self_end_thermo: Optional[float] = None
+    probe_max_hairpin_thermo: Optional[float] = None
```
tfenne, line 214: Same question here on renaming these.  FWIW while I'm pro renmaing in `Oligo` I'm about 50/50 here as this is much more specific to primer3.

prymer/primer3/primer3_parameters.py
```diff
@@ -204,6 +220,15 @@ def __post_init__(self) -> None:
...
             raise ValueError("Max threshold for dinucleotide bases must be an even number of bases")
+        default_thermo_max: float = self.probe_tms.min - 10.0
+        thermo_max_fields = [
+            "probe_max_self_any_thermo",
+            "probe_max_self_end_thermo",
+            "probe_max_hairpin_thermo",
+        ]
+        for field in fields(self):
+            if field.name in thermo_max_fields and getattr(self, field.name) is None:
+                object.__setattr__(self, field.name, default_thermo_max)
```
tfenne, line 231: Is there a reason to use getattr/setattr here instead of just having three lines that look like:

```python
self.probe_max_self_any_thermo = self.probe_max_self_any_thermo if self.probe_max_self_any_thermo is not None else default_thermo_max
```

prymer/primer3/primer3_weights.py
```diff
@@ -113,6 +120,9 @@ class ProbeWeights:
     probe_tm_gt: float = 1.0
     probe_gc_lt: float = 0.5
     probe_gc_gt: float = 0.5
+    probe_self_any_th: float = 1.0
+    probe_self_end_th: float = 1.0
+    probe_hairpin_th: float = 1.0
```
tfenne, line 125: Don't these default to 0 in primer3?  E.g. [here](https://primer3.org/manual.html#PRIMER_INTERNAL_WT_SELF_ANY_TH)

tfenne (changes requested):
Is there a reason to add the parameters and weights for internal oligos, but not for left/right primers?

prymer/primer3/primer3_parameters.py
msto: Rule of three?

Subjective preference? (The ternary is 133 characters without indentation, so it's going to get a gnarly linewrap or need a `#noqa` pragma)

A slightly simpler implementation might look like:
```py
for fieldname in thermo_max_fields:
    if getattr(self, fieldname) is None:
        object.__setattr__(self, fieldname, default_thermo_max)
```

prymer/api/oligo.py
tfenne: For me, I think this comes down to ... is this library solely about wrapping primer3, or more generally for primer design.  But I agree, that if we rename, we should also reference the primer3 return value/key/concept that matches this in the documentation.

prymer/primer3/primer3.py
tfenne: @msto My suggestion above was to make sure these values are never `None`.  There are parameters that one supplies to primer3 to turn on thermodynamic calculations, and I'm suggesting we should ensure those are always on, in which case these should never be None.

prymer/primer3/primer3_parameters.py
tfenne: Sure ... but if we're counting lines then the 5 lines do declare the list of three attributes counts too.  But more importantly, even if it's slightly more code, writing it out using symbols vs. strings is more robust to refactoring, find-usages, etc.

prymer/primer3/primer3.py
msto: Then we should also update the corresponding attributes to be non-`Optional`, correct?

prymer/api/oligo.py
msto: Is there another primer design tool you've used with `fgprimer` in the past?

prymer/api/oligo.py
```diff
@@ -103,6 +125,9 @@ class Oligo(OligoLike, Metric["Oligo"]):
     span: Span
     bases: Optional[str] = None
     tail: Optional[str] = None
+    self_any_th: Optional[float] = None
+    self_end_th: Optional[float] = None
+    hairpin_th: Optional[float] = None
```
msto, line 130: @tfenne per [this comment below](https://github.com/fulcrumgenomics/prymer/pull/60/files#r1790363446), I think these should not be `Optional`. Does that agree with what you have in mind?

prymer/primer3/primer3_parameters.py
msto, line 214: @tfenne Same question as above - do you intend for these to be non-`Optional`, if we are going to enforce that `primer3` will always return the thermodynamic results?

prymer/api/oligo.py
emmcauley: decided:
* self_any_th -> tm_homodimer
* self_end_th -> tm_3p_anchored_homodimer
* hairpin_th -> tm_secondary_structure

prymer/primer3/primer3_weights.py
emmcauley: Yes -- it was an intentional decision to deviate from the primer3 defaults, I can restore them to 0.0

prymer/primer3/primer3_parameters.py
emmcauley: I think it adds clarity to have them all be named similarly across `Oligo`, `*Weights` and `*Parameters` -- a user can tell, from the name, what parameters they are tweaking, how much each is going to be weighted, and the corresponding characteristics of the design candidate

msto: Seconding Erin here

emmcauley: I think it is consistent if these params stay optional -- if they are not provided, then we default to `probe_tm.min - 10.0` in the `__post_init__`, which is what the Primer3 manual says it does.

prymer/primer3/primer3_parameters.py
emmcauley: I'd like to mark this as resolved since we covered it in an internal meeting the other day: this is a frozen dataclass so a little extra footwork is required to set some attributes based on other attributes in a `__post_init()__`

prymer/primer3/primer3_parameters.py
ameynert: Third ;)

tfenne (approved):

PR #62, opened 2024-10-03
feat: Primer3 takes a list of VCFs instead of a VariantLookup object

### LINKED ISSUE #53: `Primer3` should accept a variant lookup `Path`
Much like `Primer3` accepts a `Path` to the input FASTA, it would be more ergonomic if the class accepted a `Path` to a VCF instead of requiring the user to construct a `VariantLookup`.

emmcauley:
Closes #53.

Summary:
* We want users to be able to provide `Primer3` with either a list of VCF paths OR a pre-constructed `VariantLookup` object if they would like to cutomize the settings (e.g., MAF filtering). 
* The choice of `VariantLookup` does not need to be exposed to users -- so this PR changes the `VariantLookup` API to such that, under the hood, the correct `VariantLookup` object is created. Depending on the value of the boolean flag `cached`, either a `_InMemoryLookup` or `_DiskBasedLookup` object is constructed. 
* The `Primer3` invocation is changed to optionally accept a list of VCF `Path`s or `VariantLookup` object (with default `cached=True`).

Tests are updated accordingly. Both the public and private objects are tested directly (so I intentionally did not address some of the comments made by coderabbit).

### DIFF
--- a/prymer/api/variant_lookup.py
+++ b/prymer/api/variant_lookup.py
@@ -1,10 +1,9 @@
 """
 # Variant Lookup Class and Methods
 
-This module contains the abstract class
-[`VariantLookup`][prymer.api.variant_lookup.VariantLookup] to facilitate retrieval of
-variants that overlap a given genomic coordinate range.  Concrete implementations must implement the
-[`query()`][prymer.api.variant_lookup.VariantLookup.query] method for retrieving variants
+This module contains the class [`VariantLookup`][prymer.api.variant_lookup.VariantLookup] to
+facilitate retrieval of variants that overlap a given genomic coordinate range.
+The [`query()`][`prymer.api.variant_lookup.VariantLookup.query`] method is used to retrieve variants
 that overlap the given range.
 
 [`VariantLookup`][prymer.api.variant_lookup.VariantLookup] needs a list of VCF files to be queried,
@@ -345,7 +344,6 @@ class VariantLookup(ContextManager):
         maf: Optional[float] = None,
         include_missing_mafs: bool = None,
     ) -> list[SimpleVariant]:
-
         maf = maf if maf is not None else self.min_maf
         include_missing_mafs = (
             include_missing_mafs if include_missing_mafs is not None else self.include_missing_mafs
--- a/prymer/api/variant_lookup.py
+++ b/prymer/api/variant_lookup.py
@@ -7,20 +7,12 @@ variants that overlap a given genomic coordinate range.  Concrete implementation
 [`query()`][prymer.api.variant_lookup.VariantLookup.query] method for retrieving variants
 that overlap the given range.
 
-Two concrete implementations are provided that both take a list of VCF files to be queried:
-
-- [`FileBasedVariantLookup`][prymer.api.variant_lookup.FileBasedVariantLookup] -- performs
-disk-based retrieval of variants (using a VCF index).  This class is recommended for large VCFs. The
-[`disk_based()`][prymer.api.variant_lookup.disk_based] alternative constructor is
-provided for easy construction of this object.
-- [`VariantOverlapDetector`][prymer.api.variant_lookup.VariantOverlapDetector] -- reads in
-variants into memory and uses an
+[`VariantLookup`][prymer.api.variant_lookup.VariantLookup] needs a list of VCF files to be queried,
+a minimum Minimum Allele Frequency with which to optionally filter variants, a boolean `include_missing_mafs` flag, and a
+boolean `cached` flag. It is recommended to set `cached` to `False` for large VCFs. When `cached`
+is True, variants from smaller VCFs are loaded into memory.
 [`pybedlite.overlap_detector.OverlapDetector`](https://pybedlite.readthedocs.io/en/latest/api.html#pybedlite.overlap_detector.OverlapDetector)
-for querying.  This class is recommended for small VCFs. The
-[`cached()`][prymer.api.variant_lookup.cached] alternative constructor is provided for
-easy construction of this object.
-
-Each class can also use minor allele frequency (MAF) to filter variants.
+is used for querying.
 
 The helper class `SimpleVariant` is included to facilitate VCF querying and reporting out results.
 
@@ -28,7 +20,7 @@ The helper class `SimpleVariant` is included to facilitate VCF querying and repo
 
 ```python
 >>> from pathlib import Path
->>> lookup = cached(vcf_paths=[Path("./tests/api/data/miniref.variants.vcf.gz")], min_maf=0.00, include_missing_mafs=True)
+>>> lookup = VariantLookup(vcf_paths=[Path("./tests/api/data/miniref.variants.vcf.gz")], min_maf=0.00, include_missing_mafs=True)
 >>> lookup.query(refname="chr2", start=7999, end=8000)
 [SimpleVariant(id='complex-variant-sv-1/1', refname='chr2', pos=8000, ref='T', alt='<DEL>', end=8000, variant_type=<VariantType.OTHER: 'OTHER'>, maf=None)]
 >>> variants = lookup.query(refname="chr2", start=7999, end=9900)
@@ -67,7 +59,6 @@ from pathlib import Path
 from types import TracebackType
 from typing import ContextManager
 from typing import Optional
-from typing import final
 
 import pysam
 from fgpyo.vcf import reader
@@ -237,7 +228,7 @@ class _VariantInterval(Interval):
         )
 
 
-class VariantLookup(ABC):
+class _VariantLookup(ABC):
     """Base class to represent a variant from a given genomic range.
 
     Attributes:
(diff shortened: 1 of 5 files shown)


tfenne:
I'd appreciate @msto's feedback on this PR.  I think original the thought was it would just be a list of VCFs vs. a VariatnLookup, but the reality of it being 4 parameters that control how variants are read from VCFs, makes it less desirable IMHO.

I think I would propose either:

- Creation of a small dataclass (e.g. `VariantParams`) to hold a list of VCFs and all the other parameters that effect VariantLookup creation, and then optionally take that in primer3.py?  (Or `VariantParams | VariantLookup | None`).
- Go back to just having it take a Variant Lookup

But I'd wait for @msto's input before doing anything further.

msto (commented):
@tfenne if it were up to me, I would drop `min_maf` and `include_missing_mafs` entirely, and defer responsibility for any allele frequency filtering to upstream preprocessing.

You could also change `Primer3` to accept a VCF path (default/no filtering applied) or a `VariantLookup`, instead of removing the latter outright, if you wanted to expose the AF parameters

prymer/api/variant_lookup.py
```diff
@@ -254,6 +250,30 @@ def __init__(
         self.min_maf: Optional[float] = min_maf
         self.include_missing_mafs: bool = include_missing_mafs
 
+    @classmethod
+    def build(
+        cls,
+        vcf_paths: list[Path],
+        use_cache: Optional[bool] = True,
+        min_maf: Optional[float] = 0.00,
+        include_missing_mafs: Optional[bool] = False,
```
msto, line 259: **suggestion** I don't think any of these need to be `Optional`. 

Two general rules of thumb - 

1. If you have a default value that isn't `None`, you probably don't need to make the parameter `Optional`
2. You shouldn't ever have an `Optional[bool]` - this gets buggy and tricky pretty quickly

🙂 

```suggestion
        use_cache: bool = True,
        min_maf: float = 0.0,
        include_missing_mafs: bool = False,
```

prymer/api/variant_lookup.py
```diff
@@ -254,6 +250,30 @@ def __init__(
...
+        for larger VCFs).
+
+        """
+        if use_cache:
+            return VariantOverlapDetector(
+                vcf_paths=vcf_paths, min_maf=min_maf, include_missing_mafs=include_missing_mafs
+            )
+        return FileBasedVariantLookup(
+            vcf_paths=vcf_paths, min_maf=min_maf, include_missing_mafs=include_missing_mafs
+        )
```
msto, line 275: **suggestion** Since the two classes have the same signature, you could make this a little shorter like so:
```py
        lookup_cls = VariantOverlapDetector if use_cache else FileBasedVariantLookup
        lookup = lookup_cls(vcf_paths=vcf_paths, min_maf=min_maf, include_missing_mafs=include_missing_mafs)
        
        return lookup
```

Otherwise I'd suggest including the `else` branch:
```py
        if use_cache:
            lookup = VariantOverlapDetector(
                vcf_paths=vcf_paths, min_maf=min_maf, include_missing_mafs=include_missing_mafs
            )
        else:
            lookup = FileBasedVariantLookup(
                vcf_paths=vcf_paths, min_maf=min_maf, include_missing_mafs=include_missing_mafs
            )

        return lookup
```

I think I've talked about this on slack, but it's easier to test and reason about when your function has a single return statement. And it's easier to drop a breakpoint and debug if the return statement doesn't also include logic

prymer/primer3/primer3.py
```diff
@@ -228,13 +227,23 @@ def __init__(
         self,
         genome_fasta: Path,
         executable: Optional[str] = None,
-        variant_lookup: Optional[VariantLookup] = None,
+        vcfs_or_lookup: Optional[list[Path] | VariantLookup] = None,
+        use_cache: bool = False,
+        min_maf: float = 0.0,
+        include_missing_mafs: bool = False,
```
msto, line 233: **issue** I think we had landed on keeping these out of the `primer3` signature.

IIRC we decided to use the `VariantLookup` default parameters if the user passed paths, and require the user to use `VariantLookup` directly if they wanted more control over these parameters.
```suggestion
```

prymer/primer3/primer3.py
```diff
@@ -228,13 +227,23 @@ def __init__(
         self,
         genome_fasta: Path,
         executable: Optional[str] = None,
-        variant_lookup: Optional[VariantLookup] = None,
+        vcfs_or_lookup: Optional[list[Path] | VariantLookup] = None,
```
msto, line 230: **question** Would it be simpler or more complicated if there were two separate arguments, `variant_vcfs` and `variant_lookup`, which were mutually exclusive?

prymer/primer3/primer3.py
```diff
@@ -244,8 +253,20 @@ def __init__(
...
+            self.vcfs_or_lookup = vcfs_or_lookup
+        elif isinstance(vcfs_or_lookup, list) and all(
+            isinstance(path, Path) for path in vcfs_or_lookup
+        ):  # if provided path, create appropriate `VariantLookup`
+            self.vcfs_or_lookup = VariantLookup.build(
+                vcf_paths=vcfs_or_lookup,
+                use_cache=use_cache,
+                min_maf=min_maf,
+                include_missing_mafs=include_missing_mafs,
+            )
```
msto, line 266: **issue** While the parameter may be a list of VCFs or a `VariantLookup`, the attribute may only be a `VariantLookup`, and we should name it and type it accordingly. 

Also, what happens if the conditions in these two branches aren't met?

```suggestion
        variant_lookup: Optional[VariantLookup]
        if vcfs_or_lookup is None or isinstance(vcfs_or_lookup, VariantLookup):
            variant_lookup = vcfs_or_lookup
        elif isinstance(vcfs_or_lookup, list) and all(
            isinstance(path, Path) for path in vcfs_or_lookup
        ):  # if provided path, create appropriate `VariantLookup`
            variant_lookup = VariantLookup.build(
                vcf_paths=vcfs_or_lookup,
                use_cache=use_cache,
                min_maf=min_maf,
                include_missing_mafs=include_missing_mafs,
            )
        
        self.variant_lookup = variant_lookup
```

msto (changes requested):

prymer/api/variant_lookup.py
tfenne: Agreed that these should be non-Optional.

Though I might argue in fact that `include_missing_mafs` should remain `Optional` and be `None` by default, with the behavior being that if `min_maf <= 0` it gets set to True, and otherwise defaults to False.

If we don't do that, then I think it should default to True - i.e. include all the variants in the VCFs until told otherwise.

prymer/api/variant_lookup.py
tfenne: I personally prefer the second (more explicit) option, and agree with @msto on using an `else` branch.

prymer/primer3/primer3.py
tfenne: Since they are mutually exclusive, I kinda like doing it this way.  Though I wonder if the type would be clearer written as `list[Path] | VariantLookup | None`?

prymer/primer3/primer3.py
```diff
@@ -228,13 +227,23 @@ def __init__(
...
             executable: string representation of the path to primer3_core
-            variant_lookup: VariantLookup object to facilitate hard-masking variants
+            vcfs_or_lookup: an optional list of VCF files or `VariantLookup` object
+                with which to hard-mask variants
+            use_cache: whether to use an in-memory `VariantLookup` for fast querying
+            min_maf: an optional minimum Minor Allele Frequency with which to filter a list of
+                VCF files (return only variants with at least this minor allele frequency)
+            include_missing_mafs: when filtering variants with a minor allele frequency,
+                `True` to include variants with no annotated minor allele frequency, otherwise
+                `False`. If no minor allele frequency is given, then this parameter does nothing.
```
tfenne, line 246:
```suggestion
```

I agree with @msto that we were going to leave these out, and users can use `VariantLookup.build()` if they want more control.

prymer/primer3/primer3.py
tfenne: I might also suggest calling this parameter `variants` or something else that can work for either the lookup or the vcfs.

prymer/primer3/primer3.py
```diff
@@ -244,8 +253,20 @@ def __init__(
             executable="primer3_core" if executable is None else executable
         )
         command: list[str] = [f"{executable_path}"]
+        if vcfs_or_lookup is None or isinstance(vcfs_or_lookup, VariantLookup):
+            self.vcfs_or_lookup = vcfs_or_lookup
+        elif isinstance(vcfs_or_lookup, list) and all(
+            isinstance(path, Path) for path in vcfs_or_lookup
+        ):  # if provided path, create appropriate `VariantLookup`
+            self.vcfs_or_lookup = VariantLookup.build(
```
tfenne, line 261: This is confusing to me - this class isn't a dataclass or similar.  There's no need to name the member property `self.vcfs_or_lookup` to match the init parameter.  Why not just call it `self.variant_lookup`?

tfenne (changes requested):

emmcauley:
closing in favor of #115

PR #64, opened 2024-10-04
feat: use OffTargetDetector and ExecutableRunner as ContextManagers

clintval:
Since `OffTargetDetector` and `ExecutableRunner` are context managers, it makes sense to type them as such by having them inherit from the abstract manager base class.

This PR also fixes an issue where if tests fail, IO resources are not cleaned up since context managers were not being used.

### DIFF
--- a/tests/api/test_picking.py
+++ b/tests/api/test_picking.py
@@ -573,31 +573,29 @@ def _pick_top_primer_pairs(
     max_primer_pair_hits: int,
     min_difference: int = 1,
 ) -> list[PrimerPair]:
-    offtarget_detector = OffTargetDetector(
-        ref=picking_ref,
-        max_primer_hits=max_primer_hits,
-        max_primer_pair_hits=max_primer_pair_hits,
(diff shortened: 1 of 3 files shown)


tests/api/test_picking.py
```diff
@@ -573,31 +573,29 @@ def _pick_top_primer_pairs(
...
+        picked: list[PrimerPair] = pick_top_primer_pairs(
+            primer_pairs=primer_pairs,
+            num_primers=len(primer_pairs),
+            min_difference=min_difference,
+            params=params,
+            offtarget_detector=offtarget_detector,
+            is_dimer_tm_ok=lambda s1, s2: (
+                dimer_checker.duplex_tm(s1=s1, s2=s2) <= params.max_dimer_tm
+            ),
+        )
```
clintval, line 597: This test helper is executed many times.

Previously, if any of these tests failed, 100s of file handles would be left open on the OS without close.

tfenne (approved):
Thanks for doing this!

PR #65, opened 2024-10-04
feat: remove pysam and bwa warmup hack

clintval:
It is mentioned `pysam.AlignmentFile` internally buffers stdin so a hack was devised to "warm it up".

However, `fgpyo.sam.reader` and `pysam.AlignmentFile` are actually not needed and only provide code indirection since we can build the SAM header and all subsequent records using the [`from_text()`](https://pysam.readthedocs.io/en/latest/api.html#pysam.AlignmentHeader.from_text)/[`fromstring()`](https://pysam.readthedocs.io/en/latest/api.html#pysam.AlignedSegment.fromstring) methods of each.

### DIFF
--- a/prymer/offtarget/bwa.py
+++ b/prymer/offtarget/bwa.py
@@ -45,10 +45,10 @@ from typing import Optional
 from typing import cast
 
 import pysam
-from fgpyo import sam
 from fgpyo import sequence
 from fgpyo.sam import Cigar
 from pysam import AlignedSegment
+from pysam import AlignmentHeader
 
 from prymer.api import coordmath
 from prymer.util.executable_runner import ExecutableRunner
@@ -201,6 +201,7 @@ class BwaAlnInteractive(ExecutableRunner):
         reverse_complement: reverse complement each query sequence before alignment.
         include_alt_hits: if True include hits to references with names ending in _alt, otherwise
                           do not include them.
+        header: the SAM alignment header.
     """
 
     def __init__(
@@ -284,20 +285,14 @@ class BwaAlnInteractive(ExecutableRunner):
 
         super().__init__(command=command)
 
-        # HACK ALERT
-        # This is a hack. By trial and error, pysam.AlignmentFile() will block reading unless
-        # there's at least a few records due to htslib wanting to read a few records for format
-        # auto-detection. Lame.  So a hundred queries are sent to the aligner to align enable the
-        # htslib auto-detection to complete, and for us to be able to read using pysam.
-        num_warmup: int = 100
-        for i in range(num_warmup):
-            query = Query(id=f"ignoreme:{i}", bases="A" * 100)
-            fastq_str = query.to_fastq(reverse_complement=self.reverse_complement)
-            self._subprocess.stdin.write(fastq_str)
-        self.__signal_bwa()  # forces the input to be sent to the underlying process.
-        self._reader = sam.reader(path=self._subprocess.stdout, file_type=sam.SamFileType.SAM)
-        for _ in range(num_warmup):
-            next(self._reader)
+        header = []
+        for line in self._subprocess.stdout:
+            if line.startswith("@"):
+                header.append(line)
+            if line.startswith("@PG"):
+                break
+
+        self.header = AlignmentHeader.from_text("".join(header))
 
     def __signal_bwa(self) -> None:
         """Signals BWA to process the queries"""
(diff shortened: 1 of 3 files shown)


prymer/offtarget/bwa.py
```diff
@@ -201,6 +201,7 @@ class BwaAlnInteractive(ExecutableRunner):
         reverse_complement: reverse complement each query sequence before alignment.
         include_alt_hits: if True include hits to references with names ending in _alt, otherwise
                           do not include them.
+        header: the SAM alignment header.
```
clintval, line 204: This adds the header as a public attribute because that seems useful.

I can revert if we don't want to make this available!

prymer/offtarget/bwa.py
```diff
@@ -412,7 +411,3 @@ def to_hits(self, rec: AlignedSegment) -> list[BwaHit]:
             hits = [hit for hit in hits if not hit.refname.endswith("_alt")]
 
         return hits
-
-    def close(self) -> None:
-        self._reader.close()
-        super().close()
```
clintval, line 418: Although we no longer need to override this method, it should have conformed to the same shape as the superclass:

```python
def close(self) -> bool:
    self._reader.close()
    return super().close()
```

prymer/offtarget/bwa.py
```diff
@@ -33,7 +33,7 @@
 >>> bwa.map_all(queries=[query])
 [BwaResult(query=Query(id='NA', bases='AAAAAA'), hit_count=3968, hits=[])]
 >>> bwa.close()
-
+True
```
clintval, line 36: Because of: https://github.com/fulcrumgenomics/prymer/pull/65/files#r1787936312

prymer/offtarget/bwa.py
```diff
@@ -284,20 +286,14 @@ def __init__(
...
-        self.__signal_bwa()  # forces the input to be sent to the underlying process.
-        self._reader = sam.reader(path=self._subprocess.stdout, file_type=sam.SamFileType.SAM)
-        for _ in range(num_warmup):
-            next(self._reader)
+        header = []
+        for line in self._subprocess.stdout:
+            if line.startswith("@"):
+                header.append(line)
+            if line.startswith("@PG"):
+                break
```
tfenne, line 294: Something about this worries me a bit.  It's probably not worth worrying about since it's unlikely `bwa aln` will ever change the header it outputs though.

prymer/offtarget/bwa.py
```diff
@@ -334,13 +330,17 @@ def map_all(self, queries: list[Query]) -> list[BwaResult]:
...
+
+        # Force the input to be sent to the underlying process.
+        self.__signal_bwa()
 
         # Read back the results
         results: list[BwaResult] = []
         for query in queries:
             # get the next alignment and convert to a result
-            results.append(self._to_result(query=query, rec=next(self._reader)))
+            line: str = next(self._subprocess.stdout).strip()
```
tfenne, line 341: I wonder if to be extra safe we should add a check here that if the line starts with `@` to keep reading lines?

tfenne (approved):

prymer/offtarget/bwa.py
clintval: @tfenne I was thinking similarly.

My first local draft of this PR sent in a single sentinel FASTQ record (similar to the original "hack") with a single base to align. The queryname would be set to `ignore-me` so we would break out of this loop if `line.startswith("ignore-me")`. Would you like this instead? It would be easy to add.

The only remaining side effect of sending in dummy/sentinel records is that bwa counts how many records it aligns so if the user was able to read the stderr then it would always seem as if there were extra records being aligned.

If we keep the current approach in this PR, I can add a unit test that asserts the shape of the dictionary. If we update `bwa-aln-interactive` to change the header, then that unit test should catch the regression.

prymer/offtarget/bwa.py
clintval: I like it! This is a good use case for `assert` since you can turn off asserts when running Python via "optimization" mode.

https://docs.python.org/3/using/cmdline.html#cmdoption-O

prymer/offtarget/bwa.py
clintval: - [ ] I should add a unit test for this

tests/offtarget/test_bwa.py
```diff
@@ -68,6 +73,19 @@ def test_hit_build_rc() -> None:
...
+    """Tests that bwa will return a properly constructed header."""
+    with BwaAlnInteractive(ref=ref_fasta, max_hits=1) as bwa:
+        header: SamHeaderType = bwa.header.to_dict()
+        assert set(header.keys()) == {"HD", "SQ", "PG"}
+        assert header["HD"] == {"GO": "query", "SO": "unsorted", "VN": "1.5"}
+        assert header["SQ"] == [{"LN": 10001, "SN": "chr1"}]
+        assert len(header["PG"]) == 1
+        program_group: dict[str, str] = cast(list[dict[str, str]], header["PG"])[0]
+        program_group["ID"] = "bwa"
+        program_group["PN"] = "bwa"
```
clintval, line 86: We should probably update `bwa-aln-interactive` to use the new executable name...

https://github.com/fulcrumgenomics/bwa-aln-interactive/issues/5

prymer/offtarget/bwa.py
clintval: I kept the current behavior but added an `assert` below to guard for leftover sequence header lines.

If bwa stops writing a program group, this code block will hang indefinitely (unit tests will catch that regression!)

clintval:
I have one strangely failing test on GitHub CI (but not locally):

```
FAILED tests/offtarget/test_bwa.py::test_header_is_properly_constructed - AssertionError: assert {'PG', 'SQ'} == {'HD', 'PG', 'SQ'}
  Extra items in the right set:
  'HD'
  Full diff:
  - {'HD', 'PG', 'SQ'}
  ?  ------
  + {'PG', 'SQ'}
```

Will debug...

tfenne (approved):

PR #69, opened 2024-10-08
fix: Check against `HN` tag instead of `is_unmapped` property

### LINKED ISSUE #68: Handling unmapped results in `BwaAlnInteractive._to_result`
`bwa aln` sometimes returns a read that is unmapped but does have hits (both primary and `XA` tag). Suggest that we ignore this case:

```
        num_hits: int = int(rec.get_tag("HN")) if rec.has_tag("HN") else 0
        if num_hits > self.max_hits:
            return BwaResult(query=query, hit_count=num_hits, hits=[])
        else:
            hits = self.to_hits(rec=rec)
            hit_count = num_hits if len(hits) == 0 else len(hits)
            return BwaResult(query=query, hit_count=hit_count, hits=hits)
```

msto:
Closes #68 

`bwa aln` produces inconsistent behavior when recording the mapped status of multiply mapping reads. In most cases, such reads are flagged as mapped with a mapq of 0. However, we have discovered some cases where these reads are flagged as unmapped. 

`prymer` currently relies on the unmapped flag to identify reads with a hit count of zero. This has been yielding errors when attempting to parse reads in the latter category. This information can instead be obtained directly from the `HN` tag.

---

NB: Thanks to detective work by @tfenne 

> This happens because internally bwa concatenates all the sequences in the FASTA file, so it can discover hits that extend across contig boundaries.  Unfortunately when it discovers it’s done this on the mapping selected as the primary mapping, it just marks it as unmapped instead of removing the mapping entirely (and selecting a different primary mapping if there are secondary hits).

This PR additionally discards any hits that extend beyond the end of the contig to which they are mapped.

### DIFF
--- a/prymer/offtarget/bwa.py
+++ b/prymer/offtarget/bwa.py
@@ -165,8 +165,10 @@ class BwaHit:
 
 @dataclass(init=True, frozen=True)
 class BwaResult:
-    """Represents zero or more hits or alignments found by BWA for the given query.  The number of
-    hits found may be more than the number of hits listed in the `hits` attribute.
+    """
+    Represents zero or more hits or alignments found by BWA for the given query.
+    
+    The number of hits found may be more than the number of hits listed in the `hits` attribute.
 
     Attributes:
         query: the query (as given, no RC'ing)
@@ -359,7 +361,8 @@ class BwaAlnInteractive(ExecutableRunner):
         return results
 
     def _to_result(self, query: Query, rec: pysam.AlignedSegment) -> BwaResult:
-        """Converts the query and alignment to a result.
+        """
+        Convert the query and alignment to a result.
 
         Args:
             query: the original query
@@ -370,17 +373,10 @@ class BwaAlnInteractive(ExecutableRunner):
                 "Query and Results are out of order" f"Query=${query.id}, Result=${rec.query_name}"
             )
 
-        num_hits: Optional[int] = int(rec.get_tag("HN")) if rec.has_tag("HN") else None
-        if rec.is_unmapped:
-            if num_hits is not None and num_hits > 0:
-                raise ValueError(f"Read was unmapped but num_hits > 0: {rec}")
-            return BwaResult(query=query, hit_count=0, hits=[])
-        elif num_hits > self.max_hits:
-            return BwaResult(query=query, hit_count=num_hits, hits=[])
-        else:
-            hits = self.to_hits(rec=rec)
-            hit_count = num_hits if len(hits) == 0 else len(hits)
-            return BwaResult(query=query, hit_count=hit_count, hits=hits)
+        num_hits: int = int(rec.get_tag("HN")) if rec.has_tag("HN") else 0
+        hits: list[BwaHit] = self.to_hits(rec=rec) if 0 < num_hits <= self.max_hits else []
+
+        return BwaResult(query=query, hit_count=num_hits, hits=hits)
 
     def to_hits(self, rec: AlignedSegment) -> list[BwaHit]:
         """Extracts the hits from the given alignment.  Beyond the current alignment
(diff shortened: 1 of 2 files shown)


prymer/offtarget/bwa.py
```diff
@@ -357,17 +360,10 @@ def _to_result(self, query: Query, rec: pysam.AlignedSegment) -> BwaResult:
...
-                raise ValueError(f"Read was unmapped but num_hits > 0: {rec}")
-            return BwaResult(query=query, hit_count=0, hits=[])
-        elif num_hits > self.max_hits:
-            return BwaResult(query=query, hit_count=num_hits, hits=[])
-        else:
-            hits = self.to_hits(rec=rec)
-            hit_count = num_hits if len(hits) == 0 else len(hits)
-            return BwaResult(query=query, hit_count=hit_count, hits=hits)
+        num_hits: int = int(rec.get_tag("HN")) if rec.has_tag("HN") else 0
+        hits: list[BwaHit] = self.to_hits(rec=rec) if 0 < num_hits <= self.max_hits else []
```
msto, line 364: Someone check my logic here - this may need to be `< self.max_hits`, because `self.max_hits` may refer to the maximum number of **secondary** hits in the XA tag (i.e. the `bwa samse -n` parameter), not the maximum number of **total** hits (which would be the number of secondary + 1)

msto: Callout from Alison - this should instead be:
- empty list when `num_hits=0`
- list containing all hits when `num_hits` is between 0 and max
- list containing only the primary alignment when `num_hits` is greater than (or equal to?) max

tbh I think this logic probably belongs in `to_hits()`, not here

msto: @tfenne Could you please weigh in on whether the list of hits should include the primary alignment or be empty when `num_hits` is greater than max?

tfenne (changes requested):
I think there needs to be a further change here, or at least a clear explanation/justification of why not.  I think when the record is marked as `unmapped`, what that really means is that `bwa` believes the primary alignment to be erroneous - and as such I think to_hits() should ignore the primary alignment and only decode the secondary alignments.

prymer/offtarget/bwa.py
```diff
@@ -357,17 +360,17 @@ def _to_result(self, query: Query, rec: pysam.AlignedSegment) -> BwaResult:
...
-            return BwaResult(query=query, hit_count=hit_count, hits=hits)
+        num_hits: int = int(rec.get_tag("HN")) if rec.has_tag("HN") else 0
+        hits: list[BwaHit] = self.to_hits(rec=rec) if 0 < num_hits < self.max_hits else []
+
+        # `to_hits()` removes artifactual hits which span the boundary between concatenated
+        # reference sequences. If we are reporting a list of hits, the number of hits should match
+        # the size of this list. Otherwise, we either have zero hits, or more than the maximum
+        # number of hits. In both of the latter cases, we have to rely on the count reported in the
+        # `HN` tag.
+        num_hits = len(hits) if hits else num_hits
```
tfenne, line 371:
```suggestion
        # `to_hits()` removes artifactual hits which span the boundary between concatenated
        # reference sequences. If we are reporting a list of hits, the number of hits should match
        # the size of this list. Otherwise, we either have zero hits, or more than the maximum
        # number of hits. In both of the latter cases, we have to rely on the count reported in the
        # `HN` tag.
        hits: list[BwaHit]
        if 0 < num_hits < self.max_hits:
            hits = self.to_hits(rec=rec)
            num_hits = len(hits)
        else:
            hits = []
```

tfenne: I think this probably handles edge cases better.  E.g. what if `HN=1` and that one hit spans a boundary, meaning you get back `[]` from `self.to_hits()`

tfenne (approved):

prymer/offtarget/bwa.py
msto: Had the same thought this morning!

PR #71, opened 2024-10-10
feat: ensure FileBasedVariantLookup is used as a context manager

clintval:
The [`FileBasedVariantLookup`](https://prymer.readthedocs.io/en/latest/reference/prymer/api/variant_lookup.html#prymer.api.variant_lookup.FileBasedVariantLookup) opens one or more [`pysam.VariantFile`](https://pysam.readthedocs.io/en/latest/api.html#pysam.VariantFile) objects but provides no public API for [closing them](https://pysam.readthedocs.io/en/latest/api.html#pysam.VariantFile.close). The file-based lookup should have a public method for closing the file handles and should also support use as a context manager so there are safe ways to cleanup IO resources.

Closes: https://github.com/fulcrumgenomics/prymer/issues/27


.github/workflows/tests.yml
```diff
@@ -35,7 +35,6 @@ jobs:
       - name: Set up miniconda
         uses: conda-incubator/setup-miniconda@v3
         with:
-          miniforge-variant: Mambaforge
```
clintval, line 38: This is necessary now, since support has dropped fully:

https://github.com/fulcrumgenomics/prymer/issues/27

Also implemented in:

- https://github.com/fulcrumgenomics/prymer/pull/32

clintval:
@coderabbitai full review

prymer/api/variant_lookup.py
```diff
@@ -353,6 +365,20 @@ def _query(self, refname: str, start: int, end: int) -> list[SimpleVariant]:
             simple_variants.extend(self.to_variants(variants, source_vcf=path))
         return sorted(simple_variants, key=lambda x: x.pos)
 
+    def close(self) -> None:
+        """Close the underlying VCF file handles."""
+        for handle in self._readers:
+            handle.close()
```
tfenne, line 371: I think that ideally the `close()` method would be defined in the ABC `VariantLookup` perhaps with a default no-op implementation.  That way it would still be callable directly by users without having to know which sub-class they are dealing with.

tfenne: And actually ... perhaps the addition of `AbstractContextManager` belongs there too, so both can be used as context managers (even though the in memory one doesn't _need_ to be)?

clintval: I try not to muddy a base class with implementations that the inherited classes don't need since that also can introduce user confusion: "_do I need to `close()` my in-memory variant lookup?_".

How should we recommend users interact with these objects?

The "alternative constructors" [`cached`](https://prymer.readthedocs.io/en/latest/reference/prymer/index.html#prymer.api.cached) and [`disk_based`](https://prymer.readthedocs.io/en/latest/reference/prymer/index.html#prymer.api.disk_based), which I think we are recommending people use, both have specific return types instead of the generic `VariantLookup`. So the user will always know if they are dealing with a pure variant lookup vs one that is file-based, deals with filesystem IO, and needs to be cleaned up.

But if we go with what's being developed here:

- https://github.com/fulcrumgenomics/prymer/pull/62/files#diff-05b5d47908e2c1c869d9e117f92fe667ca3ab1a2c25fa8839b8e94275071c710R253-R275

Then I do think we'll need to provide a generic `close()` since only a `VariantLookup` will be returned from the "alternate constructor".

clintval: I can wait to address this PR until #62 is finished and merged...

tfenne: My 2c is that ideally the two concrete classes become _private_ classes either as part of #62 or shortly thereafter, and users (including `primer3.py`) only ever construct them through `VariantLookup.build()`.  Hence the request to make `VariantLookup` the context manager with the `close()` method.

clintval: Ok! I've made the change in anticipation of that API break.

msto: Just a quick 2c - private classes really don't make sense in Python. `build` can be typed as returning `VariantLookup`, but the object the user receives will be one of the concrete classes. 

e.g. 

```py
In [4]: from abc import ABC
   ...:
   ...: class VariantLookup(ABC):
   ...:     pass
   ...:
   ...:     @classmethod
   ...:     def build(cls, cached: bool) -> "VariantLookup":
   ...:         if cached:
   ...:             return InMemoryLookup()
   ...:         else:
   ...:             return FileBasedLookup()
   ...:
   ...:
   ...: class FileBasedLookup(VariantLookup):
   ...:     pass
   ...:
   ...: class InMemoryLookup(VariantLookup):
   ...:     pass
   ...:

In [5]: lookup = VariantLookup.build(cached=False)

In [6]: type(lookup)
Out[6]: __main__.FileBasedLookup

In [7]: lookup = VariantLookup.build(cached=True)

In [8]: type(lookup)
Out[8]: __main__.InMemoryLookup
```

If you really want to hide the implementation details then the cached and disk-based logic needs to be contained within the `VariantLookup` class (which should then not be abstract)

msto: FWIW, as I type this I am thinking of two other points in favor of abandoning the current ABC pattern and abstracting this logic within `VariantLookup`

1. I would find it confusing to receive an undocumented/"private" class as proposed. 
2. It's also rather unpythonic to require construction through a custom classmethod and prevent construction through `__init__` - I would expect `VariantLookup(...)` to be able to construct the object I need

msto: A minimal change might look something like the following (truncated for brevity and clarity):

```
class VariantLookup:
    def __init__(...):
        ...
        if cached:
            self._lookup = InMemoryLookup()
        else:
            self._lookup = DiskBasedLookup()
            
    def query(...):
        return self._lookup.query(...)

class InMemoryLookup:
    ...

class DiskBasedLookup:
    ...
```

Mostly similar, but the different lookups are never exposed to the user.

You could have a private ABC, e.g. `_VariantLookup`, which both could subclass, if you wanted to maintain that class hierarchy

clintval: @msto I'd prefer that approach over what is developing in #62.

Already, I'm not a fan of adding methods to an abstract superclass that are non-sensical for some of the concrete implementations.

I don't know if we need a fancy solution. I'm also OK with having two concrete implementations that conform to a very simple "protocol" or abstract base class such that the user will pick the one they need based on the data they have on hand. Making a common entrypoint/constructor may feel convenient now, but as codebases tend to grow, so does the complexity of the primary entrypoint/constructor and you can end up in situations where you want to add parameters that are specific to one of the concrete implementations but not the others. Not saying that will happen here, but can happen in the same way that this PR evolved to attach methods on an ABC that one of the children simply doesn't need.

I have no skin in the game on this particular API design decision, and we could move the discussion to #62 or the GitHub issue that motivated it. I'm only here hoping to solve the IO cleanup issue so that I stop experiencing OS issues due to too many filehandles left unopened when tests are interrupted 😅.

msto (approved):

prymer/api/variant_lookup.py
msto: My other 2c is that this thread can and should be moved to a separate issue and follow-up PR. I think this PR satisfactorily addresses the issue at hand.

clintval: To keep things trim, I reverted the decision to move the context manager API to the parent class. This PR now let's you close the file-based variant lookup and doesn't preclude us from moving that API to a parent class in the future (or other API breaks).

tfenne (approved):
I think this is a step in the right direction.  I still think we need to refactor at some point so that if you have a `VariantLookup` you can call `close()` on it.

msto:
@tfenne I've mentioned your last comment in a new issue related to resolving some of the discussion that occurred ont his PR

https://github.com/fulcrumgenomics/prymer/issues/88

Issue #41, opened 2024-09-23
There should be a way to view the logs that bwa emits


clintval:
Interaction with BWA is implemented using [`ExecutableRunner`](https://github.com/fulcrumgenomics/prymer/blob/70f684086feb5f204e1f8fd659e7b7443cd4e845/prymer/util/executable_runner.py).

By default, all IO channels are set to `PIPE`:

https://github.com/fulcrumgenomics/prymer/blob/70f684086feb5f204e1f8fd659e7b7443cd4e845/prymer/util/executable_runner.py#L42-L44

BWA doesn't override any of them.

https://github.com/fulcrumgenomics/prymer/blob/70f684086feb5f204e1f8fd659e7b7443cd4e845/prymer/offtarget/bwa.py#L285

To interact with the `PIPE`s BWA will read and write to the private subprocess:

https://github.com/fulcrumgenomics/prymer/blob/70f684086feb5f204e1f8fd659e7b7443cd4e845/prymer/offtarget/bwa.py#L296

https://github.com/fulcrumgenomics/prymer/blob/70f684086feb5f204e1f8fd659e7b7443cd4e845/prymer/offtarget/bwa.py#L298

I think there is a small memory use bug here in that stderr will be written to a pipe that will never be read from. A buffer may be created and only increase in size. Regardless, I think that stderr should be redirected to a logging stream (perhaps with `DEBUG` level) so that a user has a way of reading them.

Originally raised in: https://github.com/fulcrumgenomics/prymer/pull/32#issuecomment-2362821429

clintval:
Sister issue for Primer3:

- https://github.com/fulcrumgenomics/prymer/issues/15

PR #73, opened 2024-10-17
fix: ensure ExecutableRunner stderr goes somewhere by default

clintval:
This doesn't fully solve:

- https://github.com/fulcrumgenomics/prymer/issues/41

But it should solve the issue where the stderr PIPE's buffer may become full and deadlock the process.


clintval:
No longer needed due to merging:

- https://github.com/fulcrumgenomics/prymer/pull/74

Issue #38, opened 2024-09-23
Remove `build_and_pick_primer_pairs()` and `pick_top_primer_pairs()`


msto:
Per @tfenne , these should only be used as an example of how to write a primer picking algorithm.

>  I will just keep repeating that I would **not** use `build_and_pick_primer_pairs` directly, especially when also designing probes, but in general too.  I would use it as an example, and place to start from, but IMHO that code should not be in `prymer` other than as example code.  The reason being that how you pick the pairs tends to vary quite a bit depending on application.

These functions should be deprecated and then removed from the package. They can be migrated to any client repo (there should be only one 🙂 ) that depends on this specific implementation. 

The contents of `pick_top_primer_pairs()` may be added to the `prymer` documentation as example code, but it should not remain in the API to avoid misuse.

PR #75, opened 2024-10-24
Reducing `picking.py` to be simpler, more focused, and more performant.

tfenne:
Closes https://github.com/fulcrumgenomics/prymer/issues/38

### DIFF
--- a/prymer/api/picking.py
+++ b/prymer/api/picking.py
@@ -120,9 +120,6 @@ def build_primer_pairs(
     Returns:
         an iterator over all the valid primer pairs, unsorted
     """
-    print(f"Target={target}; lefts={left_primers}.")
-    print(f"Target={target}; rights={right_primers}.")
-
     # Short circuit if we have no left primers or no right primers
     if not any(left_primers) or not any(right_primers):
         return
--- a/prymer/api/picking.py
+++ b/prymer/api/picking.py
@@ -130,14 +130,17 @@ def build_primer_pairs(
     Returns:
         an iterator over all the valid primer pairs, unsorted
     """
+    print(f"Target={target}; lefts={left_primers}.")
+    print(f"Target={target}; rights={right_primers}.")
+
     # Short circuit if we have no left primers or no right primers
     if not any(left_primers) or not any(right_primers):
         return
 
-    if not all(p.span.refname == target.refname for p in left_primers):
+    if any(p.span.refname != target.refname for p in left_primers):
         raise ValueError("Left primers exist on different reference than target.")
 
-    if not all(p.span.refname == target.refname for p in right_primers):
+    if any(p.span.refname != target.refname for p in right_primers):
         raise ValueError("Right primers exist on different reference than target.")
 
     # Grab the sequence we'll use to fill in the amplicon sequence
--- a/prymer/api/picking.py
+++ b/prymer/api/picking.py
@@ -4,8 +4,7 @@
 Typically, the first step is to build primer pairs from a list of left and right primers for a
 given target with the [`build_primer_pairs()`][prymer.api.picking.build_primer_pairs]
 method. The returned primer pairs are automatically scored (using the
-[`score()`][prymer.api.picking.score] method), and returned sorted by the penalty
-(increasing).
+[`score()`][prymer.api.picking.score] method).
 
 Next, the list of primer pairs for a given target are filtered based on various criteria using the
 [`pick_top_primer_pairs()`][prymer.api.picking.pick_top_primer_pairs] method. Criteria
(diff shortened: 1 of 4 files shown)


tfenne:
I'm still actively working on this - mostly writing tests and fixing up documentation.

prymer/api/picking.py
```diff
@@ -186,285 +79,108 @@ def score(
...
-    primer_pairs: list[PrimerPair] = []
-    for left_primer in left_primers:
-        for right_primer in right_primers:
-            if left_primer.span.refname != right_primer.span.refname:
-                raise ValueError(
-                    "Cannot create a primer pair from left and right primers on different"
-                    f" references: left: '{left_primer.span.refname}'"
-                    f" right: '{right_primer.span.refname}'"
+    # Short circuit if we have no left primers or no right primers
+    if not any(left_primers) or not any(right_primers):
```
msto, line 134: **issue** If either `left_primers` or `right_primers` is a generator, `any()` will consume an element.

**suggestion** Consider converting the primers to a `list`, or find a solution to peek at the first element without consuming it.

prymer/api/picking.py
```diff
@@ -186,285 +79,108 @@ def score(
...
+        # generate all the primer pairs that don't violate hard size and Tm constraints
+        for lp in left_primers:
+            for rp in right_primers:
+                if rp.span.end - lp.span.start + 1 > amplicon_sizes.max:
+                    continue
+
+                amp_mapping = Span(refname=target.refname, start=lp.span.start, end=rp.span.end)
+                amp_bases = bases[
+                    amp_mapping.start - region_start : amp_mapping.end - region_start + 1
+                ]
```
msto, line 158: **question** Is this replicating functionality that is already available via `PrimerPair`? I'm surprised that amplicon length, span, and bases wouldn't be available through that API.

prymer/api/picking.py
```diff
@@ -186,285 +79,108 @@ def score(
...
+                amp_tm = calculate_long_seq_tm(amp_bases)
```
msto, line 159: **suggestion** Related to the last comment, should this be a `cached_property` on `PrimerPair`?

prymer/api/picking.py
```diff
@@ -186,285 +79,108 @@ def score(
...
+
+                if amp_tm < amplicon_tms.min or amp_tm > amplicon_tms.max:
+                    continue
+
+                if max_heterodimer_tm is not None:
+                    if ntthal.duplex_tm(lp.bases, rp.bases) > max_heterodimer_tm:
+                        continue
+
+                penalty = score(
```
msto, line 168: **question** Naive question - why is `duplex_tm()` a method on `NtThermoAligner`, but `score()` and `calculate_long_seq_tm()` are standalone functions? Are the latter accessible through `primer3`?

msto (changes requested):

prymer/api/picking.py
tfenne: Ugh, I always forget the `Iterator` in python is `Iterable`.

prymer/api/picking.py
tfenne: `duplex_tm()` is a complicated thing that does a thermodynamic alignment of two sequences and predicts their Tm, which is why we delegate to ntthal instead of trying to replicate that.

`score()` is mimicking something in primer3, but is not easily accessible independent of designing primer pairs in primer3.

`calculate_long_seq_tm()` is so simple it's not worth trying to redirect to primer3.

prymer/api/picking.py
```diff
@@ -186,285 +79,109 @@ def score(
...
+    # Grab the sequence we'll use to fill in the amplicon sequence
+    with FastaFile(f"{fasta_path}") as fasta:
+        region_start = min(p.span.start for p in left_primers)
+        region_end = max(p.span.end for p in right_primers)
+        bases = fasta.fetch(target.refname, region_start - 1, region_end)
+
+    with NtThermoAlign() as ntthal:
+        # generate all the primer pairs that don't violate hard size and Tm constraints
+        for lp in left_primers:
+            for rp in right_primers:
```
clintval, line 152: `itertools.product` would help you un-nest this a level:

https://docs.python.org/3/library/itertools.html#itertools.product

prymer/api/picking.py
tfenne: Not really ... PrimerPair does have a function to make the amplicon mapping from the two primers, but it's trivial and also it's not a static/class method in PrimerPair :/

PrimerPair does not have any functionality to fetch the bases for the amplicon - it expected to be given it.

prymer/api/picking.py
tfenne: I'd prefer not ... we have on implementation of long sequence Tm, but it's possible other versions might be used.

prymer/api/picking.py
```diff
@@ -186,285 +69,112 @@ def score(
...
+    amplicon_tms: MinOptMax[float],
+    max_heterodimer_tm: Optional[float],
+    weights: PrimerAndAmpliconWeights,
+    fasta_path: Path,
+) -> Iterator[PrimerPair]:
     """Builds primer pairs from individual left and primers.
 
+    Only primer pairs that meet the requirements for amplicon sizes and tms will be returned.
+    In addition, if `max_heterodimer_tm` is provided then primer pairs with a heterodimer tm
+    exceeding the maximum will also be discarded.
```
msto, line 107: **question** Why is this necessary when `primer3` will apply these constraints to designed primer pairs if they are specified? e.g. specifying `PRIMER_PRODUCT_MIN_TM` and `PRIMER_PRODUCT_MAX_TM` will constrain the pairs returned by `primer3` to the provided range

msto (commented):
I don't have any objections to the refactoring.

I do continue to struggle to understand the motivation for reimplementing primer pair picking and scoring in a less performant language, rather than using `primer3` to design and filter pairs where possible.

I understand that `primer3` does not support all the possible flavors of filtering we sometimes have to implement (e.g. I don't think its design algorithm supports filtering based on heterodimerization tm), but I think that it's generally simpler and less prone to inconsistency to rely on it for the features it does provide.

prymer/api/picking.py
tfenne: Because this is used when designing left and right primers independently - so primer3 is not emitting pairs and not cannot check these constraints.

PR #79, opened 2024-11-07
Make sure the number_probes_returned is respected.

tfenne:


prymer/primer3/primer3_parameters.py
```diff
@@ -219,6 +219,7 @@ def to_input_tags(self) -> dict[Primer3InputTag, Any]:
             Primer3InputTag.PRIMER_INTERNAL_MAX_GC: self.probe_gcs.max,
             Primer3InputTag.PRIMER_INTERNAL_MAX_POLY_X: self.probe_max_polyX,
             Primer3InputTag.PRIMER_INTERNAL_MAX_NS_ACCEPTED: self.probe_max_Ns,
+            Primer3InputTag.PRIMER_NUM_RETURN: self.number_probes_return
```
clintval, line 222: How hard would it be to implement a unit test to catch a potential regression in the future?

msto, line 222: **note** `probe_max_dinuc_bases` is also not represented in the input tag mapping

msto (changes requested):
Hey @tfenne 

Thanks for catching this and pushing a fix. I noticed there's another attribute on the parameters dataclass (`probe_max_dinuc_bases`) that isn't propagated to the input tag dict.

Should we review all the parameter dataclasses to ensure that all included parameters are represented in the corresponding input tag dicts?

Also, how did you and Nils determine which subset of primer3 input tags to expose in `fgprimer` and `prymer`? I noticed there are several tags relevant to probe design which we don't include. e.g. related to your comments yesterday about ensuring the melting temperature calculation receives the same parameters as IDT uses, `prymer` does not currently permit the user to specify `PRIMER_INTERNAL_SALT_MONOVALENT` or `PRIMER_INTERNAL_SALT_DIVALENT`, or DMSO/DNA/formamide concentration

Screenshot 2024-11-08 at 9 19 17 AM

msto:
FYI `primer_max_dinuc_bases` is also not included in the input tags for `PrimerAndAmpliconParameters`. I don't see any other examples of attributes that are included on one of the `Parameters` dataclasses but not in the class's input tags dict.

I still don't know the motivation for limiting these dataclasses to a subset of the available tags/params - it looks like `prymer` does include all of the primer3 input tags in the `Primer3InputTag` enum - why not make them available for the user to specify?

prymer/primer3/primer3_parameters.py
tfenne: This one is actually on purpose - primer3 doesn't do dinuc checking.  It's definitely slightly confusing that we then supply the parameter is something that makes it look like it's going to primer3, but the check is done separately after primer3.

tfenne: I agree this would be nice, but there is not an existing test to amend, and I don't have the time to _also_ fix the test shortcomings at this moment.

PR #86, opened 2024-11-13
Use Sequence instead of list[] in parameters to build_primer_pairs()

### LINKED ISSUE #84: Update `build_primer_pairs()` to accept `Iterable[Oligo]` or `Sequence[Oligo]`
I have a use case where I have subclassed `Oligo` in order to store QC attributes (e.g. hairpin delta G and off-target hit counts) and filter individual primers before assembling them into pairs.

Because `list` is invariant, the current implementation does not support building pairs from subclasses of `Oligo`.

Note that the implementation prior to #75 supported this use case because the arguments were defined as `Iterable[Oligo]`.

tfenne:
Fixes #84

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -155,10 +155,3 @@ dmypy.json
 
 # Cython debug symbols
 cython_debug/
-
-# PyCharm
-#  JetBrains specific template is maintained in a separate JetBrains.gitignore that can
-#  be found at https://github.com/github/gitignore/blob/main/Global/JetBrains.gitignore
-#  and can be added to the global gitignore or merged into this file.  For a more nuclear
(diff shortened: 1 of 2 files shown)


msto (approved):
I don't see any reason not to implement coderabbit's suggestion here, I never would have caught that

tfenne:
> I don't see any reason not to implement coderabbit's suggestion here, I never would have caught that

I went the other way - the JetBrains section at the bottom was commented out and not that useful, and I think it's nice to have the ignores for all the IDEs in the same place.

msto:
Makes sense to me.

I was more commenting that I never would have noticed the duplication - I'm pretty lenient about additions to the gitignore 😅

PR #87, opened 2024-11-14
Emit primer pairs in penalty order.

tfenne:
This PR does two things:

1. Fix a bug where amplicons that were too _small_ were still emitted
2. Ensure that primer pairs are emitted in penalty order

(2) requires materializing a small tuple for all _valid_ pairs, and then sorting by score.  The tuple contains two ints (indices into the primer sequences) and two floats (the penalty and the tm, the last for convenience so we don't have to recompute it).  It the sorts the tuples by penalty, and starts generating PrimerPairs in penalty order.

If you have e.g. 500 left and 500 right primers, this could construct ~250k tuples and calculate 250k Tms, but in reality the number is probably substantially smaller constrained by amplicon sizes.

### DIFF
--- a/prymer/api/picking.py
+++ b/prymer/api/picking.py
@@ -121,7 +121,8 @@ def build_primer_pairs(  # noqa: C901
         fasta_path: the path to the FASTA file from which the amplicon sequence will be retrieved.
 
     Returns:
-        an iterator over all the valid primer pairs, sorted by primer pair penalty
+        An iterator over all the valid primer pairs, sorted by primer pair penalty. 
+        Primer pairs with smaller penalties are returned first.
     """
     # Short circuit if we have no left primers or no right primers
     if not any(left_primers) or not any(right_primers):
--- a/prymer/api/picking.py
+++ b/prymer/api/picking.py
@@ -121,7 +121,7 @@ def build_primer_pairs(  # noqa: C901
         fasta_path: the path to the FASTA file from which the amplicon sequence will be retrieved.
 
     Returns:
-        an iterator over all the valid primer pairs, unsorted
+        an iterator over all the valid primer pairs, sorted by primer pair penalty
     """
     # Short circuit if we have no left primers or no right primers
     if not any(left_primers) or not any(right_primers):
--- a/prymer/api/picking.py
+++ b/prymer/api/picking.py
@@ -133,6 +133,10 @@ def build_primer_pairs(  # noqa: C901
     if any(p.span.refname != target.refname for p in right_primers):
         raise ValueError("Right primers exist on different reference than target.")
 
+    # Sort the left and right primers
+    left_primers = sorted(left_primers, key=lambda p: p.span.start)
+    right_primers = sorted(right_primers, key=lambda p: p.span.end)
+
     # Grab the sequence we'll use to fill in the amplicon sequence
     with FastaFile(f"{fasta_path}") as fasta:
         region_start = min(p.span.start for p in left_primers)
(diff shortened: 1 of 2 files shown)


msto:
I also think a PR or issue to address (3) could include discussion of whether to implement Jason's heap solution to the problem, or to stick with the naive solution.

https://fulcrumgenomics.slack.com/archives/C089VCMJR/p1731593914631879?thread_ts=1731539769.443999&cid=C089VCMJR

And I don't think discussion related to (2) or (3) should block the fix of (1).

prymer/api/picking.py
```diff
@@ -136,42 +139,60 @@ def build_primer_pairs(
         region_end = max(p.span.end for p in right_primers)
         bases = fasta.fetch(target.refname, region_start - 1, region_end)
 
+    # Each tuple is left_idx, right_idx, penalty, tm
+    pairings: list[Tuple[int, int, float, float]] = []
+
+    for i in range(0, len(left_primers)):
+        for j in range(0, len(right_primers)):
+            lp = left_primers[i]
+            rp = right_primers[j]
```
msto, line 148: **suggestion** You could save a few lines and a layer of indentation in the remaining body by using `itertools.product`

```suggestion
    for lp, rp in itertools.product(left_primers, right_primers):
```

prymer/api/picking.py
```diff
@@ -136,42 +139,60 @@ def build_primer_pairs(
...
+                amplicon=amp_mapping,
+                amplicon_tm=amp_tm,
+                amplicon_sizes=amplicon_sizes,
+                amplicon_tms=amplicon_tms,
+                weights=weights,
+            )
+
+            pairings.append((i, j, penalty, amp_tm))
+
+    pairings.sort(key=lambda tup: tup[2])
```
msto, line 177: **suggestion** Add a comment indicating what we're sorting by (or refactor to use a dataclass)

msto: **issue** With this change, the docstring is stale.

> Returns:
>    an iterator over all the valid primer pairs, **unsorted**

prymer/api/picking.py
```diff
@@ -131,50 +133,89 @@ def build_primer_pairs(
...
+                left_primer=lp,
+                right_primer=rp,
+                amplicon=amp_span,
+                amplicon_tm=amp_tm,
+                amplicon_sizes=amplicon_sizes,
+                amplicon_tms=amplicon_tms,
+                weights=weights,
+            )
+
+            pairings.append((i, j, penalty, amp_tm))
```
msto, line 197: **suggestion** You might consider constructing `PrimerPair` here, and then using `dataclasses.replace` to populate `bases` below (where the `PrimerPair` is currently being instantiated).

That way the function is using established data structures from the library instead of tuples, the sort can reference `pair.penalty` instead of a tuple element, and I think it would save a few LOC overall.

Take it or leave it though 😄

prymer/api/picking.py
```diff
@@ -131,50 +133,89 @@ def build_primer_pairs(
...
+    #      subsequent right-primer can make a valid combination with that left primer
+    for i in range(0, len(left_primers)):
+        for j in range(first_right_primer_idx, len(right_primers)):
+            lp = left_primers[i]
+            rp = right_primers[j]
+
+            # If the right primer isn't "to the right" of the left primer, move on
+            if rp.span.start < lp.span.start or lp.span.end > rp.span.end:
+                first_right_primer_idx = max(first_right_primer_idx, j+1)
+                continue
```
msto, line 168: **issue** Could we cover these lines?

nh13 (approved):

PR #89, opened 2024-11-14
Small changes to simplify build_primer_pairs().

tfenne:


prymer/api/primer_pair.py
```diff
@@ -226,29 +230,37 @@ def __str__(self) -> str:
...
-        left primer to the end of the right primer.
+        Calculates the amplicon Span from the left and right primers.
+
+        Args:
+            left_primer: the left primer for the amplicon
+            right_primer: the right primer for the amplicon
+
+        Returns:
+            a Span starting at the first base of the left primer and ending at the last base of
+             the right primer
```
msto, line 244: **suggestion** Enumerate the cases where this function will raise an exception.

```suggestion
            A Span starting at the first base of the left primer and ending at the last base of
            the right primer.
            
        Raises:
            ValueError: If `left_primer` and `right_primer` have different reference names.
            ValueError: If `left_primer` is not to the left of`right_primer`. The two primers may overlap, but the left primer must start before the first base of the right primer, and end before the last base of the right primer.
```

(NB: Please run ruff to wrap the line above)

msto: **question** I've written the docstring above to reflect the behavior of the current implementation, but _are_ the primers allowed to overlap?

prymer/api/primer_pair.py
```diff
@@ -84,7 +84,11 @@ class PrimerPair(OligoLike):
     def __post_init__(self) -> None:
         # Derive the amplicon from the left and right primers. This must be done before
         # calling super() as `PrimerLike.id` depends on the amplicon being set
-        object.__setattr__(self, "_amplicon", self._calculate_amplicon())
+        object.__setattr__(
+            self,
+            "_amplicon",
+            PrimerPair.calculate_amplicon_span(self.left_primer, self.right_primer),
+        )
```
msto, line 91: **suggestion** Is this a good opportunity to refactor to make `amplicon` a `cached_property`, rather than mucking about with a private attribute and `setattr`?

e.g.

```py
@cached_property
def amplicon(self) -> Span:
    """The interval spanned by the pair's amplicon."""
    return self.calculate_amplicon_span(self.left_primer, self.right_primer)

File: prymer/api/picking.py [hunk 3]
```diff
@@ -140,13 +141,12 @@ def build_primer_pairs(
...
+                amp_span = PrimerPair.calculate_amplicon_span(lp, rp)
+
+                if amp_span.length > amplicon_sizes.max:
                     continue
 
-                amp_mapping = Span(refname=target.refname, start=lp.span.start, end=rp.span.end)
-                amp_bases = bases[
-                    amp_mapping.start - region_start : amp_mapping.end - region_start + 1
-                ]
+                amp_bases = bases[amp_span.start - region_start : amp_span.end - region_start + 1]
```
msto, line 149: **issue/question** `Span` is one-based inclusive, does the start position need to be offset by 1 to correctly index into the zero-based string?

msto (changes requested):

prymer/api/picking.py
tfenne: It works out because `region_start` is also 1-based (we subtract 1 from it when fetching from the fasta above).  So `amp_span.start - region.start` becomes a 0-based offset into the string.

prymer/api/primer_pair.py
```diff
@@ -226,29 +230,37 @@ def __str__(self) -> str:
...
-                f"right primer ref: {self.right_primer.span.refname}"
+                "Left and right primers are on different references. "
+                f"Left primer ref: {left_primer.span.refname}. "
+                f"Right primer ref: {right_primer.span.refname}"
             )
 
-        # Require that the left primer does not start to the right of the right primer
-        if self.left_primer.span.start > self.right_primer.span.end:
+        # Require that the left primer starts before the right primer
+        if left_primer.span.start > right_primer.span.start:
```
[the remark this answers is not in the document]
tfenne: We probably should.  I'll add it.

prymer/api/picking.py
msto: Could we leave a comment to that effect so it's clearer for future readers?

prymer/api/primer_pair.py
tfenne: It would be weird, because the more they overlap the more you're likely to make primer dimer, but I see no reason the code should forbid it.

prymer/api/picking.py
tfenne: Done.

prymer/api/primer_pair.py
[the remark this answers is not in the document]
tfenne: Done.

prymer/api/primer_pair.py
```diff
@@ -226,29 +230,37 @@ def __str__(self) -> str:
...
-                "Left primer start must be less than or equal to right primer end; received "
-                "left primer genome span: {self.left_primer.span}, "
-                "right primer genome span: {self.right_primer.span}"
+                "Left primer does not start before the right primer. "
+                f"Left primer span: {left_primer.span}, "
+                f"Right primer span: {right_primer.span}"
             )
 
-        return replace(self.left_primer.span, end=self.right_primer.span.end)
+        return Span(left_primer.span.refname, left_primer.span.start, right_primer.span.end)
```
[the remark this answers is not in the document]
tfenne: It would .. but I don't fine the `.span` bothersome, and I think it more clearly represents the "a primer _has_ a genomic location" rather than "a primer _is_ a genomic location".

prymer/api/primer_pair.py
msto: I'd really like to remove `_amplicon` as a "private" field.

As a consequence of the current implementation, both `asdict()` and `fields()` return a field that isn't accepted by the class constructor. This is unusual behavior and requires the user to remember and manually remove the field.

Could we make `amplicon` a `cached_property`, and update the post-init checks to reference that property? That protects all the cases I'm aware of motivating the current implementation:
1. Mutation-free access to an `amplicon` "attribute".
2. The value of `amplicon` is still derived from the input primers.

As a bonus, it removes the need to use `setattr` (currently necessary because we're trying to mutate a frozen dataclass after instantiation).

tests/api/test_primer_pair.py
```diff
@@ -556,3 +554,24 @@ def test_primer_pair_compare(
...
+
+    left = Oligo(name="l", bases="AACCGGTTAA", tm=60, penalty=1, span=Span("chr1", 150, 159))
+    right = Oligo(name="l", bases="AACCGGTTAA", tm=60, penalty=1, span=Span("chr1", 50, 59))
+    with pytest.raises(ValueError, match="Left primer does not start before the right primer"):
+        PrimerPair.calculate_amplicon_span(left, right)
+
+    left = Oligo(name="l", bases="AACCGGTTAAACGTT", tm=60, penalty=1, span=Span("chr1", 150, 164))
+    right = Oligo(name="l", bases="AACCGGTTAA", tm=60, penalty=1, span=Span("chr1", 150, 159))
+    with pytest.raises(ValueError, match="Right primer ends before left primer ends"):
+        PrimerPair.calculate_amplicon_span(left, right)
```
msto, line 577: **note** I usually try to use `pytest.mark.parametrize` to parallelize test execution as much as possible. 

Tests exit after the first failed assertion, so if you bundle test cases in a single test, you run the risk of obscuring later failures. If you parametrize, you hit as many failures as possible in a single test run, and don't find yourself stuck in a loop of addressing one test case, running the test suite, and discovering a previously hidden failure.

I usually have one test parametrized over "good" cases, with each test case paired with its expected output, and one test parametrized over "bad" cases, where each case is expected to raise an exception.

Since here the `span` is the primary variable being manipulated by each case, I'd write something like the following:

```py
@pytest.mark.parametrize(
    "left_primer,right_primer,expected_amplicon",
    [
        (Span("chr1", 50, 59), Span("chr1", 150, 159), Span("chr1", 50, 159)),
    ]
)
def test_calculate_amplicon_span(left_primer: Span, right_primer: Span, expected_amplicon: Span) -> None:
    actual_amplicon: Span = calculate_amplicon_span(left_primer, right_primer)
    assert actual_amplicon == expected_amplicon

@pytest.mark.parametrize(
    "left_primer,right_primer,error_msg"
    [(Span("chr2", 50, 59), Span("chr3", 150, 159), "different references")]
def test_calculate_amplicon_span_raises(left_primer: Span, right_primer: Span, error_msg: str) -> None:
    with pytest.raises(ValueError, match=error_msg):
        calculate_amplicon_span(left_primer, right_primer)
```

prymer/api/primer_pair.py
tfenne: I tried ... but failed.  Changed it to a cached property and tried calling it from post_init() to run the validation.  Now I get this in tests, which is super unhelpful and I'm not sure why:

```
ERROR tests/api/test_primer_pair.py - TypeError: No '__dict__' attribute on 'PrimerPair' instance to cache 'amplicon' property.
ERROR tests/api/test_primer_pair.py - TypeError: No '__dict__' attribute on 'PrimerPair' instance to cache 'amplicon' property.```

Going to merge without making the change, and we can circle back to it separately.

tests/api/test_primer_pair.py
tfenne: This is, to me, a stylistic thing, and I come down very strongly on the other side.  I almost never use `pytest.mark.parametrize` because I personally find that it makes tests less maintainable, harder to read etc.  

If the tests aren't as simple as these, then I tend to break them up into multiple test functions.

PR #90, opened 2024-11-15
Add check for amplicon size too small.

tfenne:
Once #89 is merged, this PR should re-target `main`.

### DIFF
--- a/prymer/api/picking.py
+++ b/prymer/api/picking.py
@@ -143,7 +143,8 @@ def build_primer_pairs(
             for rp in right_primers:
                 amp_span = PrimerPair.calculate_amplicon_span(lp, rp)
 
-                if amp_span.length > amplicon_sizes.max:
+                # Ignore pairings with amplicon sizes out of the range specified
+                if not amplicon_sizes.min <= amp_span.length <= amplicon_sizes.max:
                     continue
 
                 # Since the amplicon span and the region_start are both 1-based, the minuend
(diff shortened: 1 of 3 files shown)


prymer/api/picking.py
```diff
@@ -143,7 +143,8 @@ def build_primer_pairs(
             for rp in right_primers:
                 amp_span = PrimerPair.calculate_amplicon_span(lp, rp)
 
-                if amp_span.length > amplicon_sizes.max:
+                # Ignore pairings with amplicon sizes out of the range specified
+                if not amplicon_sizes.min <= amp_span.length <= amplicon_sizes.max:
```
msto, line 147: **note** In the future we could add a `.contains(x: T) -> bool` method or similar to `MinMaxOpt[T]`, which returns `True` if `x` is within the bounds of the min and max.

msto (approved):

tests/api/test_picking.py
```diff
@@ -292,6 +292,7 @@ def test_build_primers_amplicon_size_filtering(
         assert pair.amplicon.end == pair.right_primer.span.end
         assert pair.bases == REF_BASES[pair.amplicon.start - 1 : pair.amplicon.end]
         assert pair.span.length <= 200
+        assert pair.span.length >= 100
```
msto, line 295: **suggestion** Add an explicit test on the cardinality of `pairs`, rather than requiring it be non-empty?

I think we can calculate the number of pairs we expect to receive from `build_primer_pairs()`.

prymer/api/picking.py
tfenne: This is a nice idea.  I'll log an issue.

tests/api/test_picking.py
tfenne: I'd prefer not to do this ... mostly because it would make the test more brittle, and any change to the test would require either re-calculating by hand (or worse, running the test and then just copying the number over as expected).

Issue #57, opened 2024-10-02
remove `PrimerLike` as a base class


emmcauley:

emmcauley:
* conceptually linked to #57

PR #93, opened 2024-11-18
refactor: Make `PrimerPair.amplicon` a cached property

msto:
See this discussion for context: https://github.com/fulcrumgenomics/prymer/pull/89#discussion_r1847087921

This PR also updates the mypy config to exclude the `mkdocs` directories.

### DIFF
--- a/prymer/api/primer_pair.py
+++ b/prymer/api/primer_pair.py
@@ -42,8 +42,8 @@ Span(refname='chr1', start=21, end=100, strand=<Strand.POSITIVE: '+'>)
 """  # noqa: E501
 
 from dataclasses import dataclass
-from dataclasses import field
 from dataclasses import replace
+from functools import cached_property
 from typing import Iterator
 from typing import Optional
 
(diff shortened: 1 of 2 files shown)


prymer/api/primer_pair.py
```diff
@@ -79,22 +79,11 @@ class PrimerPair(OligoLike):
     amplicon_tm: float
     penalty: float
     amplicon_sequence: Optional[str] = None
-    _amplicon: Span = field(init=False)
 
-    def __post_init__(self) -> None:
```
tfenne, line 84: The problem here is that the validation that the LP/RP are oriented sensibly with respect to each other now won't run until/unless a user accesses the amplicon span.

tfenne (changes requested):

prymer/api/primer_pair.py
msto: tl;dr The validation logic is still called, and further refactoring would be related to #57

tfenne: Ugh, ok.  So a potential foot-gun to be dealt with as soon as we deal with PrimerLike.

tfenne (approved):

PR #97, opened 2024-12-05
chore: unpin fgpyo dependency

msto:
Does `fgpyo` need to be pinned to `0.7.1`, or can we use our usual [caret dependency](https://python-poetry.org/docs/dependency-specification/#caret-requirements) to permit any `0.x` minor version at 0.8 or later? 

I'd at least like to bump the version to 0.8.0 - it'd be helpful in a client project to have `fgpyo.io.assert_fasta_indexed()` available, which was added in 0.8.0.


msto:
I'd also really like to be able to use `MetricWriter` in this project! 😄

tfenne (approved):

PR #98, opened 2024-12-05
chore: unpin pybedlite dependency

msto:
Similar to #97 , this PR unpins the dependency on pybedlite.

I am opening two separate PRs in case there is an argument for one dependency to be pinned but not the other.


nh13 (approved):
lgtm

PR #99, opened 2024-12-05
feat: speed improvements for primer pair hit building from single primer hits

### LINKED ISSUE #94: Speed up from O(n^2) iteration over all left/right primer hits in `OffTargetDetector._to_amplicons`
`OffTargetDetector._to_amplicons` uses `itertools.product` over the left and right primer hits and evaluates each of them in a loop to identify valid left/right hit combinations for the pair.

Suggestion:

1. Get hits by primer sequence from `mappings_of`
2. Split into left and right hits
3. Group hits by refname
4. For each refname, split into left +, right -, left -, right+ hits
5. Build amplicons from left + and right - (amplicon strand +), left - and right + (amplicon strand -)

Building amplicons - dealing with only positive & negative in correct orientation and known to be on the same reference, with known amplicon strand
1. `itertools.product` over positive & negative primer hits
2. Check product size against min/max amplicon size range
3. If in range, return the product hit as positive start to negative end with the known reference name and amplicon strand

ameynert:
Closes #94 

1. `OffTargetDetector._to_amplicons` signature changes from 

```
def _to_amplicons(lefts: list[BwaHit], rights: list[BwaHit], max_len: int) -> list[Span]:
```

to

```
def _to_amplicons(
    positive_hits: list[BwaHit], negative_hits: list[BwaHit], max_len: int, strand: Strand
) -> list[Span]:
```

- Amplicons are built from a list of single primer hits on the positive strand and a list of single primer hits on the negative strand.
- New parameter `strand` is used to set the `strand` attribute on each of the returned `Span`s for the amplicon hits. 
- Validates that all positive hits are on the positive strand and all negative hits are on the negative strand.
- Validates that all hits are on the same reference (e.g. contig/chr).

2. The existing unit test for `OffTargetDetector.to_amplicons` is split into two, one where no error should be raised, and one where a `ValueError` should be raised, with expected error messages.

3. `OffTargetDetector._build_off_target_result` does not change signature.

- exits early setting `pass = False` to the returned `OffTargetResult` if there are too many hits for either the left or the right primer.
- hits are separated by reference (e.g. contig/chr), left/right, and strand (positive/negative) using defaultdict collections
- this enables passing hits that are on the same strand and in the correct relative orientation to each other to `_to_amplicons`, along with a known amplicon strand - positive when the positive hits are from the left primer of a pair and the negative hits are from the right primer, and negative in the inverse.
- the full set of left and right primer hits are still returned if `OffTargetDetector._keep_primer_spans` is `True`.

4. Unit test for `OffTargetDetector._build_off_target_result` is added.

### DIFF
--- a/prymer/offtarget/offtarget_detector.py
+++ b/prymer/offtarget/offtarget_detector.py
@@ -470,9 +470,12 @@ class OffTargetDetector(AbstractContextManager):
     def _to_amplicons(
         positive_hits: list[BwaHit], negative_hits: list[BwaHit], max_len: int, strand: Strand
     ) -> list[Span]:
-        """Takes a set of hits for one or more left primers and right primers and constructs
-        amplicon mappings anywhere a left primer hit and a right primer hit align in F/R
-        orientation up to `maxLen` apart on the same reference.  Primers may not overlap.
+        """Takes lists of positive strand hits and negative strand hits and constructs amplicon
+        mappings anywhere a positive strand hit and a negative strand hit occur where the end of
+        the negative strand hit is no more than `max_len` from the start of the positive strand
+        hit.
+
+        Primers may not overlap.
 
         Args:
             positive_hits: List of hits on the positive strand for one of the primers in the pair.
--- a/prymer/offtarget/offtarget_detector.py
+++ b/prymer/offtarget/offtarget_detector.py
@@ -373,7 +373,7 @@ class OffTargetDetector(AbstractContextManager):
         # Get the set of reference names with hits
         hits_by_refname: dict[str, PrimerPairBwaHitsBySideAndStrand] = {
             hit.refname: PrimerPairBwaHitsBySideAndStrand()
-            for hit in left_bwa_result.hits + right_bwa_result.hits
+            for hit in itertools.chain(left_bwa_result.hits, right_bwa_result.hits)
         }
 
         # Split the hits for left and right by reference name and strand
@@ -506,7 +506,9 @@ class OffTargetDetector(AbstractContextManager):
         if any(not h.negative for h in negative_hits):
             raise ValueError("Negative hits must be on the negative strand.")
 
-        refnames: set[ReferenceName] = {h.refname for h in positive_hits + negative_hits}
+        refnames: set[ReferenceName] = {
+            h.refname for h in itertools.chain(positive_hits, negative_hits)
+        }
         if len(refnames) > 1:
             raise ValueError("Hits are present on more than one reference.")
 
--- a/prymer/offtarget/offtarget_detector.py
+++ b/prymer/offtarget/offtarget_detector.py
@@ -362,6 +362,7 @@ class OffTargetDetector(AbstractContextManager):
             or right_bwa_result.hit_count > self._max_primer_hits
         ):
             result = OffTargetResult(primer_pair=primer_pair, passes=False)
+            return result
 
         # Get the set of reference names with hits
         hits_by_refname: dict[str, PrimerPairBwaHitsBySideAndStrand] = {
--- a/prymer/offtarget/offtarget_detector.py
+++ b/prymer/offtarget/offtarget_detector.py
@@ -504,20 +504,49 @@ class OffTargetDetector(AbstractContextManager):
         if len(positive_hits) == 0 or len(negative_hits) == 0:
             return []
 
+        # Sort the positive strand hits by start position and the negative strand hits by *end*
+        # position. The `max_len` cutoff is based on negative_hit.end - positive_hit.start + 1.
+        positive_hits_sorted = sorted(positive_hits, key=lambda h: h.start)
+        negative_hits_sorted = sorted(negative_hits, key=lambda h: h.end)
+
         amplicons: list[Span] = []
-        for positive_hit, negative_hit in itertools.product(positive_hits, negative_hits):
-            if (
-                negative_hit.start > positive_hit.end
-                and negative_hit.end - positive_hit.start + 1 <= max_len
+
+        # Track the position of the previously examined negative hit.
+        prev_negative_hit_index = 0
+        for positive_hit in positive_hits_sorted:
+            # Check only negative hits starting with the previously examined one.
+            for negative_hit_index, negative_hit in enumerate(
+                negative_hits_sorted[prev_negative_hit_index:],
+                start=prev_negative_hit_index,
             ):
-                amplicons.append(
-                    Span(
-                        refname=positive_hit.refname,
-                        start=positive_hit.start,
-                        end=negative_hit.end,
-                        strand=strand,
+                # TODO: Consider allowing overlapping positive and negative hits.
+                if (
+                    negative_hit.start > positive_hit.end
+                    and negative_hit.end - positive_hit.start + 1 <= max_len
+                ):
+                    # If the negative hit starts to the right of the positive hit, and the amplicon
+                    # length is <= max_len, add it to the list of amplicon hits to be returned.
+                    amplicons.append(
+                        Span(
+                            refname=positive_hit.refname,
+                            start=positive_hit.start,
+                            end=negative_hit.end,
+                            strand=strand,
+                        )
                     )
-                )
+
+                if negative_hit.end - positive_hit.start + 1 > max_len:
+                    # Stop searching for negative hits to pair with this positive hit.
+                    # All subsequence negative hits will have amplicon length > max_len
+                    break
+
+                if negative_hit.end < positive_hit.start:
+                    # This positive hit is genomically right of the current negative hit.
+                    # All subsequent positive hits will also be genomically right of this negative
+                    # hit, so we should start at the one after this. If this index is past the end
+                    # of the list, the slice `negative_hits_sorted[prev_negative_hit_index:]` will
+                    # be empty.
+                    prev_negative_hit_index = negative_hit_index + 1
 
         return amplicons
 
--- a/prymer/offtarget/offtarget_detector.py
+++ b/prymer/offtarget/offtarget_detector.py
@@ -75,6 +75,7 @@ BwaHit(refname='chr1', start=61, negative=True, cigar=Cigar(elements=(CigarEleme
 """  # noqa: E501
 
 import itertools
+from collections import defaultdict
 from contextlib import AbstractContextManager
 from dataclasses import dataclass
 from dataclasses import field
@@ -129,24 +130,6 @@ class OffTargetResult:
     spans: list[Span] = field(default_factory=list)
     left_primer_spans: list[Span] = field(default_factory=list)
     right_primer_spans: list[Span] = field(default_factory=list)
-
-
-@dataclass(init=True)
-class PrimerPairBwaHitsBySideAndStrand:
-    """A helper class for storing BWA hits for the left and right primers of a primer pair organized
-    by left/right and hit strand.
-
-    Attributes:
-        left_positive: A list of BwaHit objects on the positive strand for the left primer.
-        left_negative: A list of BwaHit objects on the negative strand for the left primer.
-        right_positive: A list of BwaHit objects on the positive strand for the right primer.
-        right_negative: A list of BwaHit objects on the negative strand for the right primer.
-    """
-
-    left_positive: list[BwaHit] = field(default_factory=list)
-    left_negative: list[BwaHit] = field(default_factory=list)
-    right_positive: list[BwaHit] = field(default_factory=list)
-    right_negative: list[BwaHit] = field(default_factory=list)
 
 
 class OffTargetDetector(AbstractContextManager):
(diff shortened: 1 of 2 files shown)


prymer/offtarget/offtarget_detector.py
```diff
@@ -334,26 +353,70 @@ def _build_off_target_result(
...
-                primer_pair=primer_pair,
-                passes=self._min_primer_pair_hits <= len(amplicons) <= self._max_primer_pair_hits,
-                spans=amplicons if self._keep_spans else [],
-                left_primer_spans=(
-                    [self._hit_to_span(h) for h in p1.hits] if self._keep_primer_spans else []
-                ),
-                right_primer_spans=(
-                    [self._hit_to_span(h) for h in p2.hits] if self._keep_primer_spans else []
-                ),
+
```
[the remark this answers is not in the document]
ameynert: Fixed!

prymer/offtarget/offtarget_detector.py
```diff
@@ -334,27 +353,74 @@ def _build_off_target_result(
...
-        if p1.hit_count > self._max_primer_hits or p2.hit_count > self._max_primer_hits:
+        left_bwa_result: BwaResult = hits_by_primer[primer_pair.left_primer.bases]
+        right_bwa_result: BwaResult = hits_by_primer[primer_pair.right_primer.bases]
+
+        # If there are too many hits, this primer pair will not pass. Exit early.
+        if (
+            left_bwa_result.hit_count > self._max_primer_hits
+            or right_bwa_result.hit_count > self._max_primer_hits
+        ):
             result = OffTargetResult(primer_pair=primer_pair, passes=False)
```
tfenne, line 364: Could we just return here?  Is there any value to caching the result in this case, where it's so little work to compute the result?

prymer/offtarget/offtarget_detector.py
```diff
@@ -334,27 +353,74 @@ def _build_off_target_result(
...
-                passes=self._min_primer_pair_hits <= len(amplicons) <= self._max_primer_pair_hits,
-                spans=amplicons if self._keep_spans else [],
-                left_primer_spans=(
-                    [self._hit_to_span(h) for h in p1.hits] if self._keep_primer_spans else []
-                ),
-                right_primer_spans=(
-                    [self._hit_to_span(h) for h in p2.hits] if self._keep_primer_spans else []
-                ),
+            if self._cache_results:
+                self._primer_pair_cache[primer_pair] = replace(result, cached=True)
```
tfenne, line 366: If keeping this (and again on line 408), why not just set `cached=self._cache_results` at construction time, so there's no need to `replace(cached=True`?

prymer/offtarget/offtarget_detector.py
```diff
@@ -334,27 +353,74 @@ def _build_off_target_result(
...
+            return result
+
+        # Get the set of reference names with hits
+        hits_by_refname: dict[str, PrimerPairBwaHitsBySideAndStrand] = {
+            hit.refname: PrimerPairBwaHitsBySideAndStrand()
+            for hit in left_bwa_result.hits + right_bwa_result.hits
```
tfenne, line 372: These feels pretty inefficient .. it will concatenate two fairly large lists _and_ (I think?) generate a `PrimerPairBwaHitsBySideAndStrand` per hit, then unique them, rather than generate one per reference contig.

I think something like this would work:

```python
for refname in set(hit.refname for hit in itertools.chain(left_bwa_result.hits, right_bwa_result.hits)
```

prymer/offtarget/offtarget_detector.py
```diff
@@ -334,27 +353,74 @@ def _build_off_target_result(
...
+            if hit.negative:
+                hits_by_refname[hit.refname].left_negative.append(hit)
+            else:
+                hits_by_refname[hit.refname].left_positive.append(hit)
+
+        for hit in right_bwa_result.hits:
+            if hit.negative:
+                hits_by_refname[hit.refname].right_negative.append(hit)
+            else:
+                hits_by_refname[hit.refname].right_positive.append(hit)
```
tfenne, line 386: Now that I see this ... I think actually maybe you don't care any more about left vs. right here and beyond this point?  Sorry, as I think I may have sent you down this path.

Specifically if a single left primer happens to bind to the genome in two places, close together, in the right relative orientations, that could still cause problems.  So I think it's really just one collection of + strand mappings and one collection of - strand mappings.

prymer/offtarget/offtarget_detector.py
```diff
@@ -334,27 +353,74 @@ def _build_off_target_result(
...
+
+        # Build amplicons from hits on the same reference with valid relative orientation
+        amplicons: list[Span] = []
+        for hits in hits_by_refname.values():
+            amplicons.extend(
+                self._to_amplicons(
+                    positive_hits=hits.left_positive,
+                    negative_hits=hits.right_negative,
+                    max_len=self._max_amplicon_size,
+                    strand=Strand.POSITIVE,
```
tfenne, line 396: I'm not sure how to think about strand here.  I think it's probably mostly irrelevant and could just always be set to positive?

prymer/offtarget/offtarget_detector.py
```diff
@@ -410,19 +476,50 @@ def mappings_of(self, primers: list[PrimerType]) -> dict[str, BwaResult]:
...
-                continue
+        if any(h.negative for h in positive_hits):
+            raise ValueError("Positive hits must be on the positive strand.")
+        if any(not h.negative for h in negative_hits):
+            raise ValueError("Negative hits must be on the negative strand.")
 
-            plus, minus = (h2, h1) if h1.negative else (h1, h2)
-            if minus.start > plus.end and (minus.end - plus.start + 1) <= max_len:
-                amplicons.append(Span(refname=plus.refname, start=plus.start, end=minus.end))
+        refnames: set[str] = {h.refname for h in positive_hits + negative_hits}
```
tfenne, line 505:
```suggestion
        refnames: set[str] = {h.refname for h in itertools.chain(positive_hits, 
 negative_hits)}
```

prymer/offtarget/offtarget_detector.py
```diff
@@ -410,19 +476,50 @@ def mappings_of(self, primers: list[PrimerType]) -> dict[str, BwaResult]:
...
+        if len(refnames) > 1:
+            raise ValueError("Hits are present on more than one reference.")
```
tfenne, line 507: Include the refnames here?

prymer/offtarget/offtarget_detector.py
```diff
@@ -410,19 +476,50 @@ def mappings_of(self, primers: list[PrimerType]) -> dict[str, BwaResult]:
...
+
+        amplicons: list[Span] = []
+        for positive_hit, negative_hit in itertools.product(positive_hits, negative_hits):
+            if (
+                negative_hit.start > positive_hit.end
+                and (negative_hit.end - positive_hit.start + 1) <= max_len
```
tfenne, line 513: Should we allow for situations where the two primers' mappings overlap on the genome?  E.g. maybe the check should be:

```python
negative_hit.start > positive_hit.start and negative_hit.end > positive_hit.end
```

prymer/offtarget/offtarget_detector.py
```diff
@@ -410,19 +476,50 @@ def mappings_of(self, primers: list[PrimerType]) -> dict[str, BwaResult]:
...
 
-            plus, minus = (h2, h1) if h1.negative else (h1, h2)
-            if minus.start > plus.end and (minus.end - plus.start + 1) <= max_len:
-                amplicons.append(Span(refname=plus.refname, start=plus.start, end=minus.end))
+        refnames: set[str] = {h.refname for h in positive_hits + negative_hits}
+        if len(refnames) > 1:
+            raise ValueError("Hits are present on more than one reference.")
+
+        amplicons: list[Span] = []
+        for positive_hit, negative_hit in itertools.product(positive_hits, negative_hits):
```
tfenne, line 510: I think this is good enough if it meets your performance needs @ameynert.  As noted elsewhere, if you had many many hits to one reference, you could speed this up by sorting the positive and negative strand hits, looping over list indices, and avoiding whole swathes of pairs.

tfenne (changes requested):

prymer/offtarget/offtarget_detector.py
ameynert: Caching is required by the unit test at https://github.com/fulcrumgenomics/prymer/blob/2656adf87e814c80111f7b6272ce38787119a4dc/tests/offtarget/test_offtarget.py#L99

prymer/offtarget/offtarget_detector.py
ameynert: I think the logic is that the returned object should have `cached=False`, because it's the first time it's been generated, and the cached object has `cached=True` for when it's retrieved from the cache. This jives with the description of `cached` at https://github.com/fulcrumgenomics/prymer/blob/2656adf87e814c80111f7b6272ce38787119a4dc/prymer/offtarget/offtarget_detector.py#L111

```
        cached: True if this result is part of a cache, False otherwise.  This is useful for testing
```

I suggest we come back to the cache issue on a subsequent PR.

prymer/offtarget/offtarget_detector.py
ameynert: That seems reasonable. Do you think we'd ever see a situation where a primer is a palindrome and hits the same target both forward and reverse?

prymer/offtarget/offtarget_detector.py
ameynert: In the `PrimerPair` code, it's not explicitly enforced by a `__post_init__` but is strongly assumed by e.g. [calculate_amplicon_span](https://github.com/fulcrumgenomics/prymer/blob/2656adf87e814c80111f7b6272ce38787119a4dc/prymer/api/primer_pair.py#L223C9-L223C32) that the left primer is left of the right primer. This implies that the left primer is on the positive strand and the right primer is on the negative strand. All amplicons are coded as on the positive strand (default value for `Span`, again, not explicit).

When a hit to a primer pair is in the same orientation as the primer pair definition, e.g. left positive and right negative, then it makes sense for the amplicon to also be on the positive strand.

If the hit is in the opposite orientation to the primer pair definition, the amplicon should be on the negative strand.

I think it makes sense to assign strandedness like this:

- left primer + / right primer - = amplicon +
- left primer + / left primer - = amplicon +
- right primer + / left primer - = amplicon -
- right primer + / right primer - = amplicon -

prymer/offtarget/offtarget_detector.py
ameynert: This change causes a number of tests to fail. Since it's a breaking change, I'd like to address it in a separate PR.

```
FAILED prymer/offtarget/offtarget_detector.py::prymer.offtarget.offtarget_detector
FAILED tests/offtarget/test_offtarget.py::test_ok[True] - AssertionError: assert 1187 == 856
FAILED tests/offtarget/test_offtarget.py::test_ok[False] - AssertionError: assert 1187 == 856
FAILED tests/offtarget/test_offtarget.py::test_check_too_many_primer_pair_hits[at the maximum-204-856-True-True] - AssertionError: at the maximum
FAILED tests/offtarget/test_offtarget.py::test_check_too_many_primer_pair_hits[at the maximum-204-856-True-False] - AssertionError: at the maximum
FAILED tests/offtarget/test_offtarget.py::test_to_amplicons[No mappings - overlapping primers (1bp overlap)-positive0-negative0-+-expected0-True] - AssertionError: No mappings - overlapping primers (1bp overlap)
FAILED tests/offtarget/test_offtarget.py::test_to_amplicons[No mappings - overlapping primers (1bp overlap)-positive0-negative0-+-expected0-False] - AssertionError: No mappings - overlapping primers (1bp overlap)
FAILED tests/offtarget/test_offtarget.py::test_to_amplicons[No mappings - amplicon size too big (1bp too big)-positive1-negative1-+-expected1-True] - AssertionError: No mappings - amplicon size too big (1bp too big)
FAILED tests/offtarget/test_offtarget.py::test_to_amplicons[No mappings - amplicon size too big (1bp too big)-positive1-negative1-+-expected1-False] - AssertionError: No mappings - amplicon size too big (1bp too big)
```

prymer/offtarget/offtarget_detector.py
ameynert: As discussed elsewhere, will leave questions of strand and L/L R/R hits to a separate PR.

prymer/offtarget/offtarget_detector.py
ameynert: I've tested this separately and the speed improvement by sorting the pos/neg strand hits is significant when there are a large number of hits returned, e.g. in a scenario where the max mismatches parameters are relaxed, one of the primer sequences is getting 100+ hits, and there are multiple highly similar reference contigs in the indexed genome that's being searched.

prymer/offtarget/offtarget_detector.py
tfenne: Ah, ok.  Perhaps I'll "fix" this in a separate PR.

tfenne (approved):

PR #100, opened 2024-12-06
ci: add support for building with python 3.13

nh13:


pyproject.toml
clintval, line 1: I can't comment directly but you should add 3.13 to the classifiers on line 22-23.

clintval (approved):
Could you update the badge on the README too?

It is here: https://github.com/fulcrumgenomics/prymer/blob/7874c66fc68aacf74eb20ad755952936f3fb233a/README.md?plain=1#L9

PR #101, opened 2024-12-13
feat: use primer3-py for primer3 instead of the executable

nh13:


prymer/api/oligo.py
```diff
@@ -208,7 +208,7 @@ def __str__(self) -> str:
         """
         # If the bases field is None, replace with MISSING_BASES_STRING
         bases: str = self.bases if self.bases is not None else MISSING_BASES_STRING
-        return f"{bases}\t{self.tm}\t{self.penalty}\t{self.span}"
+        return f"{bases}\t{self.tm:.2f}\t{self.penalty:.2f}\t{self.span}"
```
nh13, line 211: QoL improvement

prymer/primer3/primer3.py
```diff
@@ -399,54 +380,38 @@ def design(self, design_input: Primer3Input) -> Primer3Result:  # noqa: C901
...
-        primer3_results: dict[str, str] = {}  # key-value pairs of results reported by Primer3
+        # split the tags into sequence and non-sequence tags
+        seq_args = {}
+        global_args = {}
+        for tag, value in assembled_primer3_tags.items():
+            tag_str = f"{tag}"
+            if tag_str.startswith("SEQUENCE"):
+                seq_args[tag_str] = value
+            else:
+                global_args[tag] = value
```
tfenne, line 392: This works ... but I think we could probably also add a `is_global_arg()` to `Primer3InputTag` that does this more efficiently (e.g. based on ordinal) _or_ split `Primer3InputTag` into two classes - one for globals and one for sequence tags?

prymer/primer3/primer3.py
```diff
@@ -399,54 +380,38 @@ def design(self, design_input: Primer3Input) -> Primer3Result:  # noqa: C901
...
+        design_primers_retval = primer3.design_primers(seq_args=seq_args, global_args=global_args)
+        primer3_results: dict[str, Any] = {}  # key-value pairs of results reported by Primer3
+
+        for key, value in design_primers_retval.items():
+            # Because Primer3 will emit both the input given and the output generated, we
+            # discard the input that is echo'ed back by looking for tags (keys)
+            # that do not match any Primer3InputTag
+            if not any(key == item.value for item in Primer3InputTag):
+                primer3_results[key] = value
```
tfenne, line 401: Is this actually needed?  I don't think there's any harm in allow the inputs to be present in the dict() used below.  

If you feel strongly about this .. perhaps we have a more efficient way than looping through all the values in Primer3InputTag - e.g. create a _ALL_INPUT_TAG_STRINGS set and at the class level and check for membership?

prymer/primer3/primer3.py
```diff
@@ -399,54 +380,38 @@ def design(self, design_input: Primer3Input) -> Primer3Result:  # noqa: C901
...
             """Formats the Primer3 error and raises a ValueError."""
             error_message = f"{message}: "
             # add in any reported PRIMER_ERROR
             if "PRIMER_ERROR" in primer3_results:
                 error_message += primer3_results["PRIMER_ERROR"]
-            # add in any error lines
-            if len(error_lines) > 0:
-                error_message += "\n".join(f"\t\t{e}" for e in error_lines)
             # raise the exception now
             raise ValueError(error_message)
```
tfenne, line 410: I think this can be inline at the single call site now?

tfenne (commented):
This is very cool - and I'm surprised at actually how small of a PR this is!!

prymer/primer3/primer3.py
nh13: Since `PrimerInputTag`'s value is a string, I don't think it's inefficient, so I am going to centralize this as properties in the enum, and get rid of the f-string call (since I don't think that is needed).

prymer/primer3/primer3.py
nh13: I'll keep removing them, we can open a PR later if you want to see them.  I made it not inefficient via set membership.

prymer/primer3/primer3.py
```diff
@@ -399,55 +380,32 @@ def design(self, design_input: Primer3Input) -> Primer3Result:  # noqa: C901
...
+            # discard the input that is echo'ed back by looking for tags (keys)
+            # that do not match any Primer3InputTag
+            if key not in Primer3InputTag:
+                primer3_results[key] = value
 
         # Check for any errors.  Typically, these are in error_lines, but also the results can
         # contain the PRIMER_ERROR key.
-        if "PRIMER_ERROR" in primer3_results or len(error_lines) > 0:
-            primer3_error("Primer3 failed")
+        if "PRIMER_ERROR" in primer3_results:
```
ameynert, line 404: I'm not sure I follow what's happening here. `error_lines` is no longer present, could be removed from the comment on L402-403. Is the `else` on L407-408 unreachable?

nh13: Fixed, silly copy and paste left an extra outer `if`.  Good catch!

PR #102, opened 2025-01-01
feat: allow for left/right primer hits to overlap when building primer pair hits with `OverlapDetector`

ameynert:
See [discussion](https://github.com/fulcrumgenomics/prymer/pull/99#discussion_r1884336606).

The code before this PR does not return primer pair hits where the hits to the left and right primers overlap, and does not specify a minimum acceptable amplicon size.

This change adds two new parameters to the `OffTargetDetector` to address this:
* `allow_overlapping_hits` - defaults to `False` to maintain current behaviour. If this is set to `True`, then primer pair hits where the left and right primer hits overlap are allowed, as long as they meet the acceptable amplicon size criteria.
* `min_amplicon_size` - defaults to `1` to maintain current behaviour, and is overridden by `allow_overlapping_hits=False`.

### DIFF
--- a/prymer/api/picking.py
+++ b/prymer/api/picking.py
@@ -165,13 +165,13 @@ def build_primer_pairs(  # noqa: C901
 
             # If the right primer isn't "to the right" of the left primer, move on
             if rp.span.start < lp.span.start or lp.span.end > rp.span.end:
-                first_right_primer_idx = max(first_right_primer_idx, j+1)
+                first_right_primer_idx = max(first_right_primer_idx, j + 1)
                 continue
 
             amp_span = PrimerPair.calculate_amplicon_span(lp, rp)
 
             if amp_span.length < amplicon_sizes.min:
-                first_right_primer_idx = max(first_right_primer_idx, j+1)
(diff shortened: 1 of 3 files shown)


tfenne:
I took a quick look at this ... and I'm leaning towards thinking that we should:
a) not even parameterize, but always allow overlapping hits
b) default the "min amplicon size" to something slightly higher, e.g. 20bp

@nh13 and @ameynert what do you think of that?

ameynert:
> I took a quick look at this ... and I'm leaning towards thinking that we should: a) not even parameterize, but always allow overlapping hits b) default the "min amplicon size" to something slightly higher, e.g. 20bp
> 
> @nh13 and @ameynert what do you think of that?

I agree - I implemented it this way with the intention of not making a breaking change, but I think a user is more likely to not care too much if the primers overlap a bit so long as they get a long enough amplicon.

prymer/api/picking.py
```diff
@@ -165,13 +165,13 @@ def build_primer_pairs(  # noqa: C901
 
             # If the right primer isn't "to the right" of the left primer, move on
```
ameynert, line 166: These changes were from linting.

PR #103, opened 2025-01-07
feat: Add parameter validation for `OffTargetDetector.__init__`

ameynert:
- `max_amplicon_size > 0`
- `max_primer_hits >= 0`
- `max_primer_pair_hits >= 0`
- `min_primer_pair_hits >= 0`
- `three_prime_region_length > 0`
- `0 <= max_mismatches_in_three_prime_region <= three_prime_region_length`
- `max_mismatches >= 0`
- `max_gap_opens >= 0`
- `max_gap_extends == -1` or `max_gap_extends >= 0`

### DIFF
--- a/prymer/offtarget/offtarget_detector.py
+++ b/prymer/offtarget/offtarget_detector.py
@@ -193,20 +193,23 @@ class OffTargetDetector(AbstractContextManager):
             ref: the reference genome fasta file (must be indexed with BWA)
             max_primer_hits: the maximum number of hits an individual primer can have in the genome
                 before it is considered an invalid primer, and all primer pairs containing the
-                primer failed.
+                primer failed. Must be greater than or equal to 0.
             max_primer_pair_hits: the maximum number of amplicons a primer pair can make and be
-                considered passing
+                considered passing. Must be greater than or equal to 0.
             min_primer_pair_hits: The minimum number of amplicons a primer pair can make and be
                 considered passing. (In most cases, this is the number of amplicons a primer pair is
                 expected to generate.) The default is 1, which is appropriate when the primer pair
                 is being evaluated for off-target hits against the same reference genome from which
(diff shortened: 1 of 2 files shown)


prymer/offtarget/offtarget_detector.py
```diff
@@ -191,31 +191,83 @@ def __init__(
...
+        if min_primer_pair_hits < 0:
+            errors.append(
+                "'min_primer_pair_hits' must be greater than or equal to 0. "
+                f"Saw {min_primer_pair_hits}"
+            )
+        if three_prime_region_length < 1:
+            errors.append(
+                "'three_prime_region_length' must be greater than 0. "
+                f"Saw {three_prime_region_length}"
+            )
```
tfenne, line 253: Given that we use this to drive the "seed region" for bwa I think we should probably enforce a higher miniumum.  @nh13 do you have a thought on what's reasonable?  E.g. 8, 10, 12 or somewhere around there?

tfenne (approved):
Thanks for doing this @ameynert.  Looks great but would like us to settle on a higher min for 3' region length before merging please.  If @nh13 doesn't respond, I think we could set that to 10 for now.

ameynert:
I've added checks for `max_gap_opens` and `max_gap_extends` as well. Will wait on @nh13 to agree on a min for 3' region length before merging.

prymer/offtarget/offtarget_detector.py
nh13: I agree that since this value is use for the seed length, we should (1) add that explanation to the parameter or class docs, and (2) have a minimum of 8.

tfenne:
@ameynert as things have gotten a bit complicated, why don't we merge this with `three_prime_region_length > 0` and we can circle back and increase it as/when we see fit?

PR #106, opened 2025-01-14
Remove clustering.py.

tfenne:


nh13 (approved):
Docs failing, but take it or leave it if you want to fix that now on or later.

PR #107, opened 2025-01-14
Remove OligoLike

tfenne:
Removes the OligoLike super-class. Also:
- Removes the deprecated `Primer` subclass of Oligo
- Bumps python version to 3.12 (for now, would like to get back to 3.11 after more work)
- Removes some functions from Oligo and PrimerPair
- Converts some functions in Oligo and PrimerPair to properties and cached properties


prymer/api/oligo.py
```diff
@@ -106,58 +106,77 @@ class Oligo(OligoLike, Metric["Oligo"]):
...
     tm_3p_anchored_homodimer: Optional[float] = None
     tm_secondary_structure: Optional[float] = None
-    bases: Optional[str] = None
     tail: Optional[str] = None
 
     def __post_init__(self) -> None:
-        super(Oligo, self).__post_init__()
+        if len(self.bases) == 0:
+            raise ValueError("Bases must not be an empty string.")
+        elif self.span.length != len(self.bases):
```
nh13, line 136: Do we ever have a situation where the primer is targeting an alternate haplotype with an indel so then lengths would differ?

prymer/api/oligo.py
```diff
@@ -106,58 +106,77 @@ class Oligo(OligoLike, Metric["Oligo"]):
...
-    def longest_hp_length(self) -> int:
-        """Length of longest homopolymer in the oligo."""
-        if self.bases is None:
-            return 0
-        else:
-            return longest_homopolymer_length(self.bases)
+    @property
+    def id(self) -> str:
+        """Returns the name if there is one, otherwise the span."""
+        return self.name if self.name is not None else str(self.span)
```
nh13, line 144: Nit
```suggestion
        return self.name if self.name is not None else f"{self.span}"
        
```

prymer/api/primer_pair.py
```diff
@@ -78,8 +74,28 @@ class PrimerPair(OligoLike):
...
     penalty: float
+    name: Optional[str] = None
     amplicon_sequence: Optional[str] = None
 
+    def __post_init__(self) -> None:
+        # Force generation of the amplicon span, which validates the relative positions
+        # of the left and right primers
+        amp = self.amplicon
+
+        # If supplied, bases must be the same length as the span length
```
nh13, line 85: ditto, do we think that the amplicon supplied here matches the genome (i.e. the span) or can the amplicon be an alternative haplotype, and span is just the mapping relative to the reference (and not relative to the actual haplotype being amplified).  I suppose one could suggest that the _mapping_ should be relative to the haplotype instead of the genome.

nh13 (approved):
Minor comments, take them or leave them.

Given we have a few "TODOs" in the PRs merging to `dev`, shall we aggregate them into a list that we can add to a draft PR with `dev` merging into `main`?

prymer/api/oligo.py
tfenne: It's theoretically possible .. but I'd suggest we relax this constraint the first time that comes up, as it's more likely a bug.

prymer/api/primer_pair.py
tfenne: Yeah - I think until we or someone else tries to make a designer that targets a haplotype other than the one in the FASTA file, we should leave these checks in place (they were there previously, I just simplified).

PR #109, opened 2025-01-14
Re-organizing key dataclasses into a single `model.py` and then exposing those for import from `prymer`.

tfenne:
Instead of having a module per class, this consolidates several classes including Strand, Span, Oligo, PrimerPair etc. into a single module.  All the symbols are then re-exported so that they can be accessed via:

```python
from prymer import Span
from prymer import Oligo
...
```


prymer/api/coordmath.py
```diff
@@ -1,51 +1,7 @@
...
-# Methods for coordinate-based math and interval manipulation.
-
-Contains the following public methods:
-
-- [`require_same_refname()`][prymer.api.coordmath.require_same_refname] -- ensures that all
-    provided intervals have the same reference name.
-- [`get_locus_string()`][prymer.api.coordmath.get_locus_string] -- returns a formatted
-    string for an interval (`<refname>:<start>-<end>`).
-- [`get_closed_end()`][prymer.api.coordmath.get_closed_end] -- gets the closed end of an
-    interval given its start and length.
```
nh13, line 11: Shall we keep this doc?

prymer/api/picking.py
```diff
@@ -26,10 +26,10 @@
...
 
 from prymer.api.melting import calculate_long_seq_tm
-from prymer.api.minoptmax import MinOptMax
-from prymer.api.oligo import Oligo
-from prymer.api.primer_pair import PrimerPair
-from prymer.api.span import Span
+from prymer.model import MinOptMax
+from prymer.model import Oligo
+from prymer.model import PrimerPair
+from prymer.model import Span
```
nh13, line 32: Shall we prefer importing from the top-level, here and elsewhere?  Or is there a reason for the preference to import it from model?
```suggestion
from prymer import MinOptMax
from prymer import Oligo
from prymer import PrimerPair
from prymer import Span
```

prymer/offtarget/offtarget_detector.py
```diff
@@ -11,7 +11,7 @@
 
 ```python
 >>> from pathlib import Path
->>> from prymer.api.span import Strand
+>>> from prymer.model import Strand
```
nh13, line 14: I think in the docs we should import these as we would recommend users to import them, namely from the top-level prymer namespace.  If you agree, can you fix in the rest of the modules as well?

```suggestion
>>> from prymer import Strand
```

nh13 (approved):
Approving assuming that the imports `from prymer.model` will get changed to `from prymer`.

tfenne:
Answering here @nh13 rather than to individual comments.  On the imports - it is unfortunately necessary that within the `prymer` codebase itself that we use imports from the module where the code _is_ rather than where it's re-exported to.  If we were to e.g. use `from prymer import Span` within `prymer` you end up very quickly with circular imports where e.g. `prymer.primer3` wants to use `prymer.Span` which requires initializing `prymer` ... which also e.g. re-exports symbols from `prymer.primer3`, etc. etc.

That said, all examples and doc-tests can be updated to use the re-exported location(s) and I've done that.  Additionally tests can import from the top level too, and I have also updated those.

PR #112, opened 2025-01-15
Merge Primer*Weights into Primer3Parameters, remove Primer3Input

nh13:
1. merges the `Primer*Weights` classes into their corresponding amplicon and probe parameter classes (reduces 4 classes to two).  The attributes have the `_wt` suffix in their name.  The `primer3.primer3.primer3_weights` module is gone.
2. moves the `Primer3Input` functionality into `Primer3` directly.  `primer3.primer3.primer3_input` is now gone.
3. Adds in `WeightRange` for the weights when a value is out of range (either less than, or greater than).  Could use a better name?
4. `Primer3.design` now explicitly takes in the design task and target, in addition to the params.  It checks to make sure that the task (e.g. design left primers) matches the params (e.g. amplicons).
5. Simplify `Primer3Task` (remove methods no longer used)


prymer/primer3/primer3_parameters.py
```diff
@@ -62,19 +62,61 @@ class stores user input for internal probe design and maps it to the correct Pri
...
 
-from prymer.model import MinOptMax
+from prymer.model import MinOptMax, Span
 from prymer.primer3.primer3_input_tag import Primer3InputTag
+from prymer.primer3.primer3_task import Primer3TaskType
+
+
+class Primer3Parameters(ABC):
+    target: Span
+    task: Primer3TaskType
```
tfenne, line 79: I think I would prefer to keep these separate and just replace `Primer3Input` with providing 3 args to `primer3.design()` so that the same parameters can be re-used across many calls for many targets without having to `dataclasses.replace(param, target=next_target)` each time.

prymer/primer3/primer3_parameters.py
```diff
@@ -93,6 +135,26 @@ class PrimerAndAmpliconParameters:
         primer_max_3p_homodimer_tm: the max melting temperature acceptable for self-complementarity
             anchored at the 3' end
         primer_max_hairpin_tm: the max melting temperature acceptable for secondary structure
+        product_size_lt: weight for products shorter than
+            `PrimerAndAmpliconParameters.amplicon_sizes.opt`
+        product_size_gt: weight for products longer than
+            `PrimerAndAmpliconParameters.amplicon_sizes.opt`
```
tfenne, line 141: The frequency with which these come in pairs makes me think maybe we'd want to have a `WeightPair` (better name needed) class like `MinOptMax` so that you could do:

```python
amplicon_size_wts = Weights(0.5, 1.5)
```

tfenne: Also two other thoughts:
- I think the weights should match naming with above, so e.g. where we've said "amplicon_tms" we should use "amplicon" instead of "product" in the weights (or vice versa).
- I think for clarity all weights should either have a `wt_` prefix or `_wt` suffix; some do and some don't here.

prymer/primer3/primer3_parameters.py
```diff
@@ -105,7 +167,8 @@ class PrimerAndAmpliconParameters:
     melting temperature threshold (i.e. when provided, values should be specified independent
     of primer design.)
     """
-
+    target: Span
+    task: Primer3TaskType
```
tfenne, line 171: Since these are in the super class, do you need to repeat them here?

prymer/primer3/primer3_parameters.py
```diff
@@ -62,19 +62,61 @@ class stores user input for internal probe design and maps it to the correct Pri
...
+        """Use this method when you want to treat these parameters as amplicon parameters."""
+        if isinstance(self, AmpliconParameters):
+            return typing.cast(AmpliconParameters, self)
+        raise Exception("The parameters are not amplicon parameters")
+
+    def as_probe_params(self) -> "ProbeParameters":
+        """Use this method when you want to treat these parameters as probe parameters."""
+        if isinstance(self, ProbeParameters):
+            return typing.cast(ProbeParameters, self)
+        raise Exception("The parameters are not amplicon parameters")
```
tfenne, line 91: This feels out of place to me - why not just check and cast where that's needed rather than having it in the superclass?

prymer/primer3/primer3.py
```diff
@@ -310,21 +311,21 @@ def _screen_pair_results(
...
+                > design_input.primer_max_dinuc_bases
             ):  # if the right primer has too many dinucleotide bases, fail it
                 dinuc_pair_failures.append(primer_pair.right_primer)
                 valid = False
             if valid:  # if neither failed, append the pair to a list of valid designs
                 valid_primer_pair_designs.append(primer_pair)
         return valid_primer_pair_designs, dinuc_pair_failures
 
-    def design(self, design_input: Primer3Input) -> Primer3Result:  # noqa: C901
+    def design(self, design_input: Primer3Parameters) -> Primer3Result:  # noqa: C901
```
tfenne, line 328: I think this would need to take:

```python
amplicon_parameters: AmpliconParameters | None, probe_parameter: ProbeParameters | None
```

What you have works for the current set of use cases, but one primer3 task we haven't implemented yet supports designing primer pairs and probes concurrently, and then you need both sets of parameters.

nh13: I leave it for now, and we can decide later if we want to have two parameters to `design`, or to create a third parameters class:
```python
class AmpliconAndProbeParameters(Primer3Parameters):
    amplicon_params: AmpliconParams
    probe_params: ProbeParams
```

prymer/primer3/primer3.py
```diff
@@ -731,29 +754,3 @@ def _check_design_results(design_input: Primer3Input, design_results: dict[str,
     count: int = int(maybe_count)
 
     return count
-
-
-def _has_acceptable_dinuc_run(design_input: Primer3Input, oligo_design: Oligo) -> bool:
```
nh13, line 736: I inlined this

prymer/api/picking.py
```diff
@@ -30,7 +30,7 @@
 from prymer.model import Oligo
 from prymer.model import PrimerPair
 from prymer.model import Span
-from prymer.primer3 import PrimerAndAmpliconWeights
+from prymer.primer3 import AmpliconParameters
```
tfenne, line 33: If we're going to rename it from `PrimerAndAmplicon` I'd vote for `PrimerParameters` as it feels like the majority of the parameters are around primers.  Also, can be used for designing just left primers (no amplicon), just right primers (no amplicon) or both (and therefore an amplicon).

prymer/model.py
```diff
@@ -83,6 +83,55 @@ def __str__(self) -> str:
         return f"(min:{self.min}, opt:{self.opt}, max:{self.max})"
 
 
+@dataclass(slots=True, frozen=True, init=True)
+class WeightRange(Generic[Numeric]):
```
tfenne, line 87: I like this, but a few thoughts:

- It's not really a range, more of a pair.  I.e. I can have a higher penalty weight for "primer tm too low" than "primer tm too high" - they're independent values.  Perhaps it's just `Weights`?
- I think in p3 weights are always floats because they are scaling factors, so i think we could un-generify it and juse always have the values be floats

tfenne (approved):
Some nit-picking on naming but happy to merge once that's addressed!

PR #115, opened 2025-01-17
Refactor of VariantLookup and related code

tfenne:
Tests will continue to fail until bumped to use a version of fgpyo that includes https://github.com/fulcrumgenomics/fgpyo/pull/215


prymer/primer3/primer3.py
```diff
@@ -30,7 +30,7 @@
 >>> design_region = Span(refname="chr2", start=9095, end=9120)
 >>> soft_masked, hard_masked = designer.get_design_sequences(region=design_region)
 >>> soft_masked
-'AGTTACATTACAAAAGGCAGATTTCA'
+'AGTTAcatTACAAAAGGCAGATTTCA'
```
nh13, line 33: !!!

prymer/variant.py
```diff
@@ -0,0 +1,259 @@
+import logging
```
nh13, line 1: given `VariantLookup` is public, it'd be nice to have a module-level docstring with examples.  Would it be easy to port over the examples from the deleted module?

prymer/variant.py
```diff
@@ -0,0 +1,259 @@
...
+@dataclass(slots=True, frozen=True, init=True)
+class SimpleVariant:
+    """Represents a variant from a given genomic range.
+
+    Each variant has a single reference and alt allele.  Start and end positions are both 1-based
+    inclusive.  Unlike VCF, empty alleles are allowed and positions are given as follows:
+
+    - SNPs and MNPS will have start/end set to the first and last substituted bases
+    - Deletions will have start/end set to the first and last _deleted_ base
+    - Insertions will have start/end set to the bases between which the insertion occurs
```
nh13, line 27: I get what you mean by _between_, but it just doesn't track to say "between an insertion" for me 
```suggestion
    - Insertions will have start/end set to the bases before and after which the insertion occurs respectiely
```

prymer/variant.py
```diff
@@ -0,0 +1,259 @@
...
+        """
+        if "CAF" in variant.info:
+            # Assumes CAF is a list of allele frequencies, with the first being the reference allele
+            return [float(af) for af in variant.info["CAF"][1:]]
+        elif "AF" in variant.info:
+            return [float(af) for af in variant.info["AF"]]
+        elif "AC" in variant.info and "AN" in variant.info:
+            an = float(variant.info["AN"])
+            return [float(ac) / an for ac in variant.info["AC"]]
+
```
nh13, line 136: Why not use "FORMAT/GT" if available like previously?  I was surprised to see them gone, and didn't see a comment in the review highlighting why.

prymer/variant.py
```diff
@@ -0,0 +1,259 @@
...
+        else:
+            soft = list(bases)
+            hard = list(bases)
+
+            for variant in variants:
+                for pos in range(variant.start, variant.end + 1):
+                    if region.start <= pos <= region.end:
+                        idx = pos - region.start
+                        soft[idx] = soft[idx].lower()
+                        hard[idx] = "N"
```
nh13, line 252: Can we avoid a loop as such?
```suggestion
                start = max(variant.start, region.start)
                end = min(variant.end, region.end) + 1
                soft[start:end] = soft[start:end].lower()
                hard[start:end] = "N" * (end - start + 1)
```

nh13 (approved):
Looking good, nothing blocking.  If you don't feel like creating docstring examples, can you log an issue?

prymer/variant.py
tfenne: I'll add it to our 4.0 tracker ... my thought it to iterate and get to a point where we're happy with all the APIs then spend a day on documentation and examples.

prymer/variant.py
tfenne: I worded slightly differently but close to what you suggested.

prymer/variant.py
tfenne: I went through and updated it to return the actual AF per alternate allele rather than `1 - ref_allele_freq` for _each_ alt.  It felt like that would be quite a bit more work to do accurately using GTs, and I'm not sure that's ever been used.

prymer/variant.py
tfenne: Not easily.  `soft[start:end].lower()` doesn't work because `soft[start:end]` is a list slice, not a string.  So that line has to become:

```python
soft[start:end] = [ch.lower() for ch in soft[start:end]]
```

which just moves the loop into a different spot.  I can add in the max/min calls though instead of the `if`.

PR #118, opened 2025-01-28
fix: allow newer pybwa

znorgaard:


pyproject.toml
```diff
@@ -43,7 +43,7 @@ fgpyo = {git ="https://github.com/fulcrumgenomics/fgpyo" , rev= "c5eb469a4fbe4f2
 pysam = "^0.22.1"
 ordered-set = "^4.1.0"
 primer3-py = "^2.0.3"
-pybwa = "1.3.3"
+pybwa = "^1.3.3"
 
 [tool.poetry.group.dev.dependencies]
 poetry      = "^1.8.2"
```
nh13, line 49: we don't support poetry version 2 :/

nh13 (changes requested):
I think we have to upgrade to poetry 2 in the pyproject.toml and github actions if you want this

pyproject.toml
```diff
@@ -43,7 +43,7 @@ fgpyo = {git ="https://github.com/fulcrumgenomics/fgpyo" , rev= "c5eb469a4fbe4f2
 pysam = "^0.22.1"
 ordered-set = "^4.1.0"
 primer3-py = "^2.0.3"
-pybwa = "1.3.3"
+pybwa = "^1.3.3"
```
nh13, line 46: Might as well use the latest version
```suggestion
pybwa = "^1.4.4"
```

nh13 (approved):

PR #120, opened 2025-02-12
chore: Bump `fgpyo` to 1.0

msto:


msto:
@nh13 if we merge this could it be accompanied by a 3.0.1 release?

I think a patch bump is appropriate based on this comment: https://github.com/semver/semver/issues/148#issuecomment-21970506

PR #123, opened 2025-02-12
chore: Fix publishing workflow

msto:
## Summary

Copies the fix referenced in https://github.com/rickstaa/action-contains-tag/pull/18 and implemented in pybwa

https://github.com/fulcrumgenomics/pybwa/blob/9193fb98fdebf63ed264fa08667b4dbbf79c796c/.github/workflows/publish_pybwa.yml#L16-L22


msto:
@nh13 is there an effective way to test this short of tagging another release?

nh13 (approved):

PR #125, opened 2025-02-13
chore: Bump `fgpyo` to `^1.0.0`

msto:
We have a client repository depending on the `dev` branch of `prymer`, and this pin is causing some dependency conflicts. 

The pinned commit is on the `main` branch of `fgpyo`, so I think it's safe to move to a version pin instead.


.github/workflows/tests.yml
```diff
@@ -46,7 +46,7 @@ jobs:
 
       - name: Install the project's dependencies
         shell: bash -el {0}
-        run: poetry install -v
+        run: poetry install --no-cache -v
```
msto, line 49: As of the time of this writing, `numpy=2.2.3` was released 11 minutes ago.

Locally, `poetry lock` found the updated numpy version and added it to the lock file.

However, on github actions, `poetry install` could not find the latest numpy version as it is too recent. Disabling cache seems to fix.

nh13 (approved):

msto:
Tests are currently failing because local poetry locked to numpy=2.2.3, which [came out half an hour ago](https://pypi.org/project/numpy/), and GHA poetry can't find the new version yet.

And, afaict, there's no way to give poetry a constraint at the command line (e.g. to create a lockfile with numpy at 2.22.2). 

https://github.com/python-poetry/poetry/issues/2758

We just have to wait for the cache to refresh 😞

msto:
For posterity, this thread helped me resolve the issue:

https://github.com/python-poetry/poetry/issues/8139

These failures may occur when the local lock file was created before PyPI completed generating all the wheels for a package version. Waiting a bit and then re-generating the lock file was sufficient to resolve.

PR #126, opened 2025-02-13
chore: Add verbose retry when installing dependencies

msto:
## Summary

If `poetry install` fails in the test run, try again with increased verbosity.

The retry may resolve transient errors, and if it doesn't, we get a more detailed error log.


msto:
We should probably pull this into the Python template too.

nh13 (approved):

PR #128, opened 2025-03-14
chore: add doi to README

nh13:


znorgaard (approved):
LGTM!

One thing I noticed on this PR is the readme has a different format when compared to our other repos.

Our other repos have  breaks between (non-zenodo) badge sections. Then the tool name as a header with a one sentence description. Then the zenodo-badge.

See pybedlite
image

We probably want to standardize on that.

